PR #46, opened 2015-06-09
Export to HTML now exports embedded files.

julianstirling:
Hi Giuspen

Thanks for writing CherryTree, I am starting to use it as a lab book. I really like it for this as I can give others read access to an HTML export without them needing to install anything.

One thing that was worrying though was any embedded files were not exported as part of the HTML export. I have added this functionality in my first commit. I only do it for the HTML node export, not for the HTML selection export, because as far as I can see this is only for copying to the clipboard (am I correct on this?).

I am also thinking of extending the command line functionality to allow HTML export so I can automate it.

Julian

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+*.pyc
+*.mo
+*.directory
--- a/modules/exports.py
+++ b/modules/exports.py
@@ -467,8 +467,10 @@ class Export2Html:
             new_folder += "2"
         self.new_path = os.path.join(dir_place, new_folder)
         self.images_dir = os.path.join(self.new_path, "images")
+        self.embed_dir = os.path.join(self.new_path, "EmbeddedFiles")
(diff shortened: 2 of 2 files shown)


giuspen:
Hi Julian, I will look at your work, thank you for sharing it.
You can also export to html the only selection, it is false that the
selection is only for the clipboard.
Regards,
Giuseppe.

julianstirling:
Guiseppe,

When I said selection is only called by clipboard I meant the function "selection_export_to_html" which saves images to a temporary file and is called (from my search) from clipboard.py. The export for selection calls "node_export_to_html" with an optional argument to set the selection. I just checked and exporting a selection to HTML seems to work with my edit.

Julian

giuspen:
Julian, your analysis is correct. I'll check your work soon, thanks.
Giuseppe.

PR #53, opened 2015-07-19
Command line HTML export

julianstirling:
Hi Giuseppe (sorry for misspelling your name before!)

As I said before I was planning to make HTML exporting possible from the command line, so I can automate the export. I have managed this. I just create a "Ghost window" for cherry tree which does all the initialisation, but never shows the window, then I run the full export for all nodes. I have updated the command line options, to allow an output option and an option for the export. When starting it doesn't check for a D-Bus session, because then it can't export if the program is open.

I made on other change to the HTML export for the normal graphical method. Now when the export folder already exists it numbers them
file.ctb_HTML
file.ctb_HTML002
file.ctb_HTML003
file.ctb_HTML004
instead of
file.ctb_HTML
file.ctb_HTML2
file.ctb_HTML22
file.ctb_HTML222

Julian

### DIFF
--- a/cherrytree
+++ b/cherrytree
@@ -51,6 +51,8 @@ if ARGPARSE_OK:
     parser = argparse.ArgumentParser()
     parser.add_argument('filepath', nargs='?', help='Optional CherryTree Document to Open')
     parser.add_argument("-n", "--node", help="Node Name to Focus")
+    parser.add_argument("-x", "--export", dest='mode', action='store_const', const='export', default='open', help="Export to HTML (-o must be set)")
+    parser.add_argument("-o", "--output", action='store', help="Set output directory.")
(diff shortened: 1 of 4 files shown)


giuspen:
Hi Julian,
Many thanks for sharing your work I'll look about it and merge soon.
Take care,
Giuseppe.

giuspen:
Hi Julian, I merged your work but applied a couple of changes, let me know
if you disagree with my changes.

julianstirling:
I had a skim through. Everything seems reasonable, I haven't had a chance to pull the code and test, I will soon. As long as I can export from command line I am happy, exactly how the functions are called and the commands are sent is your choice, this project is your baby.

PR #57, opened 2015-08-14
Fixed HTML unclosed tags and text overlapping edge of page.

julianstirling:
Hi Giuseppe,

A couple of bug fixes for the new style of HTML export from Emily V:
* Tree needs `</ol>` tags after last node in list.
* Tree missing `</a>` tag from top level nodes.
* Final `</div>` tag missing
* Width didn't include margin so text extended 380px to right of screen. Fixed in css by changing margin to padding and extending the bounding box to include padding.
 
I have also added the name of the current node to top of the page.

### DIFF
--- a/glade/styles.css
+++ b/glade/styles.css
@@ -43,7 +43,8 @@ div.tree {
 
 div.page {
     width: 100%;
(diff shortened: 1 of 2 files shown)


giuspen:
Great work thank you Julian!

Issue #72, opened 2015-10-08
BasKet Import: note type "file" missing


roccosportal:
When I import a basket folder, all attachments are missing. 

.basket extract

```
<note lastModification="2015-10-06T22:27:59" added="2015-10-06T22:27:59" type="file">
      <content>somepdf.pdf</content>
</note>
```

With a quick look in the source code I guess here would be the right place to implement this:

```python
"""imports.py:BasketHandler.note_parse()"""
elif note_dom_iter.attributes['type'].value == "file":
```

Imported folder: .kde/share/apps/basket
BasKet Note Pads  Version 2.10-beta
Using KDE Development Platform 4.14.2

PR #76, opened 2015-10-14
Possible fix for #72

roccosportal:
This should add the note type file when importing from a BasKet folder. Since it's my first time working with cherrytree I'm not sure if I missed something. Code should be tested thoroughly.


giuspen:
Thank you Rocco!

Issue #73, opened 2015-10-08
BasKet Import: note type "cross_reference" missing 


roccosportal:
When I import a basket folder, all cross references are missing. 

.basket extract

```
<note lastModification="2015-10-04T23:41:44" added="2015-10-04T23:41:39" type="cross_reference">
     <content title="Some Title" icon="basket">basket://basket18-225549/</content>
</note>
```

With a quick look in the source code I guess here would be the right place to implement this:

``` python
"""imports.py:BasketHandler.note_parse()"""
elif note_dom_iter.attributes['type'].value == "cross_reference":
```

Imported folder: .kde/share/apps/basket
BasKet Note Pads  Version 2.10-beta
Using KDE Development Platform 4.14.2

PR #77, opened 2015-10-16
possible fix for #73

roccosportal:
When importing from a BasKet folder the cross references are missing. Since there is no cross reference feature in cherrytree yet, this simple fix just adds a text in the node like "cross reference: {title-of-basket-cross-reference}". i18n is not supported yet.


giuspen:
Thank you Rocco

PR #94, opened 2016-01-22
10% of lt translation

zygimantus:


giuspen:
I wanted to release tomorrow but I can delay a bit if you plan to complete in few days

zygimantus:
OK. I will try do my best.

zygimantus:
By the way, those underscores are for short keys?

giuspen:
Yes the underscores are for short keys, not mandatory you can skip them

PR #116, opened 2016-03-13
Basic implementation of line height settings.

bfloch:
This adds an unique option to the settings/config which drives GtkSourceView/GtkTextView.set_pixels_above_lines and set_pixels_below_lines.

Solves one TODO on your list and this request: [www.giuspen.com/topic/line-height-setting/](http://www.giuspen.com/topic/line-height-setting/)

A proper implementation could be relative line-height based on the font height, but I felt that this basic implementation works well enough.

**Please note**: This adds an offset to the line numbers (which seems to be an SourceView issue as the above spacing is not being added to the numbers/gutter), but the numbers still match.
The only alternative I see is only adding the space below, but then the line highlight would be off, which I personally find more distracting.

### DIFF
--- a/modules/config.py
+++ b/modules/config.py
@@ -167,6 +167,7 @@ def config_file_load(inst):
         inst.rt_highl_curr_line = config.getboolean(section, "rt_highl_curr_line") if config.has_option(section, "rt_highl_curr_line") else True
         inst.pt_highl_curr_line = config.getboolean(section, "pt_highl_curr_line") if config.has_option(section, "pt_highl_curr_line") else True
         inst.space_around_lines = config.getint(section, "space_around_lines") if config.has_option(section, "space_around_lines") else 0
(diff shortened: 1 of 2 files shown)


bfloch:
Also worth mentioning, this does not affect code boxes.

giuspen:
Thank you for your work, I'm going to merge it in a couple of days, just
after releasing 0.36.7

bfloch:
I noticed pixels-inside-wrap needs also to be set, or it should have its own setting so people could control how packed wrapped lines should be. Will add this.

bfloch:
Added vertical spacing in wrapped lines. I decided to make it relative based on the regular vertical spacing. This way you can have wrapped lines being visually packed together while still being affected.

giuspen:
Bear with me, I will merge your work in a couple of days, thank you for your contribute

PR #118, opened 2016-03-18
Fix German translation

basicmaster:
The German translation had the same translation for cut/copy/paste plain text, which is fixed by this PR.


giuspen:
Cheers!

PR #158, opened 2016-07-05
Lithuanian translation changed

zygimantus:


giuspen:
Thanks, I see that there are still 6 fuzzy strings though when opening with Poedit

PR #210, opened 2016-11-28
Word count feature

cyingfan:
This implementation uses https://bitbucket.org/emptypage/uniseg-python and a sqlite3 library with unicode data.
The feature can be toggled in Preferences->Miscellaneous->Miscellaneous->Word Count.
The word count is done on node change and on text buffer change.


giuspen:
Hi and thanks for your work.
I am willing to merge your work but it's unlikely that I will merge in the foreign library.
I'm interested to have a word count but I'm not interested to have a too complicated one, I want a lightweight one.
The recount would happen at node change plus not at every buffer change but maybe after the user presses space or enter (and maybe after paste from clipboard as well) - see support.py
`def on_sourceview_event_after_key_press(dad, text_view, event, syntax_highl):
...
    elif keyname in [cons.STR_KEY_RETURN, cons.STR_KEY_SPACE]:
...`
I can take care to connect the refresh of the value while you can for now connect only to the node change but important is that you do not use external libraries

cyingfan:
Only capturing <space> or <enter> isn't enough. There are at least two scenarios where this would not work:

1. Typing without space after punctuation, as in "hello.bye".
2. Certain languages (e.g. Thai and Chinese) do not rely on space for word separation.

giuspen:
I'm just concerned to not slow down too much the application as you type.
This option is anyway configurable and disabled by default so it should not be a problem.
Would you be able to make it work without the need of the external library? I'm not even sure I can bundle that stuff in cherrytree.

cyingfan:
I extracted parts of the code I need from uniseg and it works fine. I am undecided between using a slim down version of the sqlite3 database or having a python dict with 27K+ entries. Will need some profiling to make the call.

cyingfan:
@giuspen Hi. Committed my latest change. Please review. Thanks

giuspen:
@cyingfan Hi, great job. There's only one thing, the following connect:
`def update_selected_node_statusbar_info(self):
#...
self.curr_buffer.connect("changed", self.on_buffer_change)
#...`
I would rather add it after the other connect in:
`def on_node_changed(self, *args):
#...
self.curr_buffer.connect('modified-changed', self.on_modified_changed)
self.curr_buffer.connect("changed", self.on_buffer_change)
#...`
Let me know if this would work for you.

giuspen:
I noticed that a string like `2016/12/05 - 23:31` counts 9 words which is quite weird, except for this seems to work properly

giuspen:
also `one-two-three` counts as five while I would expect three...

PR #227, opened 2017-02-11
patch for git diff by ljw

NiceKingWei:
Dear Mr.Giuseppe:
Thanks for writing CherryTree.I really like it and I use it to write documents and take notes.I just changed a little code so that '.ctb' files can be transformed into text files through command line parameters.It does make sense when I use 'git'.
                                                                                                                                      Yours sincerely,
                                                                                                                                                                ljw


giuspen:
Hi @NiceKingWei thanks for your work.
I could not take your patch as is but had to make it consistent with the current export to html.
I understand your need is to be able to overwrite the output folder if existing, I will add an additional command line flag to allow that working for both txh and html

PR #275, opened 2017-08-05
Fix incorrect version

leggewie:
By using $upstreamversion-1 you are polluting Debian namespace.  Please use $upstreamversion-0$anythingyouwant for your own packages.  This makes it easier to update, track bugs, etc.

In all honesty, I usually advise upstreams against releasing their own binary distro packages.  It really isn't easy to get right and usually creates a bunch of problems. (Debian Maintainer speaking here)


giuspen:
I appreciate help in the proper way of packaging. I try and release binaries myself because I don't always have prompt support from the distros packagers

PR #296, opened 2017-11-10
Optional ability to change temp folder for other OS

hugemistake:
Sometime there is a need to change the temp folder location.


giuspen:
Cheers

PR #351, opened 2018-03-16
Add config option to customize 'Insert today's note' node

ainoneko:
Added  misc.journal_day_format option (default: "%d %a") to config.cfg  to be able to change the node format


giuspen:
In config.py you did not cover the case of config.cfg missing. Fix that and then I'll merge

ainoneko:
Is it correct now?

giuspen:
Yep cheers

Issue #369, opened 2018-05-21
Does not import Zim journal


Signum:
I love the feature to import Zim wikis. Zim has worked for me for four years but apparently cannot cope with many documents so I'm going with CherryTree now.
I managed to import almost all documents. However my Zim journal was not imported – it was just skipped. Maybe the problem lies in https://github.com/giuspen/cherrytree/blob/master/modules/imports.py#L787 which does not recurse into Zim structures like Journal/2018/05/21.txt
Not even the "Journal" node is created in CherryTree.
As a workaround I imported the Journal as "text" which worked but obviously I lost all my formatting.

frenchiveruti:
I have the same issue, is there any workaround?

giuspen:
@frenchiveruti can you attach a test document that reproduces the issue or send me privately? Also specify what is not being imported in your test case

frenchiveruti:
Hello giuspen, yes, I can. 
I didn't delete the zim folder and the contents aren't private. 
I've used `Zim 0.73.5` and `CherryTree 0.99.33` respectively.  

In my test space nothing is being imported no matter what I do with the Zim folder, I had some success using markdown and pandoc and then importing with markdown.  
Here's the zip file with my Zim Folder.  
[Introducción-BD.zip](https://github.com/giuspen/cherrytree/files/6313500/Introduccion-BD.zip)

frenchiveruti:
I tried, HTML file 
HTML folder
Zim Folder
Markdown file
Markdown folder

Now I do not remember which of all of them "worked", but one of them worked.

kklepper:
Cannot import Zim according to https://giuspen.com/cherrytreemanual/#_importing

> Select a file or folder to import then click Open.
> 
> In the Who is the parent? menu, select one of the following options:
> 
> The Tree Root - Each imported file is added to a top-level node.
> 
> The Selected Node - Each imported file is added to a child node of the selected node.
>
I can open the Zim directory, but nothing happens. So I look for the `Who is the parent?`-menu item -- where the heck is that? Embarrassing.

giuspen:
I tried the document attached from @frenchiveruti and it imported successfully with the current version.
The manual is discontinued unfortunately, import pointing to the root folder of zim and if it fails you can send me the dtaa privately if you want and I can give it a try.

PR #370, opened 2018-05-21
Handle recursive Zim import / Make journal today format configurable

Signum:
Attempt to fix issue #369

### DIFF
--- a/modules/imports.py
+++ b/modules/imports.py
@@ -787,6 +787,7 @@ class ZimHandler():
     def parse_folder(self, curr_folder):
         """Start the Parsing"""
         for element in os.listdir(curr_folder):
+            path = os.path.join(curr_folder, element)
             if len(element) > 4 and element[-4:] == ".txt" \
             and os.path.isfile(os.path.join(curr_folder, element)):
                 file_descriptor = open(os.path.join(curr_folder, element), 'r')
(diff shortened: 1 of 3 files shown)


modules/imports.py
```diff
@@ -797,10 +798,13 @@ def parse_folder(self, curr_folder):
                 self.node_add(wiki_string.decode(cons.STR_UTF8), node_name, curr_folder)
                 # check if the node has children
                 children_folder = os.path.join(curr_folder, node_name)
-                if os.path.isdir(children_folder):
-                    self.parse_folder(children_folder)
```
giuspen: This code allows to recognise that if "nodename.txt" exists and also at the same level exists a folder "nodename", the .txt files inside the folder "nodename" will be direct children of the node resulting from "nodename.txt". With your code we lose this but we get that every foldername is parsed also if there is not a foldername.txt beside that. I understand your goal but I would rather add the second part without removing the first, just being careful to do not parse the same folder twice, so just excluding from the second parse a folder that also has a foldername.txt at the same level

giuspen:
I can look at this only after I finish the porting from python to C++ that is quite urgent

giuspen:
This won't be looked after in Pygtk2 version, only in future Gtkmm3 version

PR #447, opened 2019-01-08
(Port to C++) add a toolbar

txe:
Added a toolbar. 
To test: click on the first button "Add node" writes "CtApp::add_node() is called" in the application output.


giuspen:
Great job thanks, only observation is I call pVariable the normal pointers and with rVariable the glib reference pointers to make it clear you won't have to delete those, so in this case would be rGtkBuilder

txe:
BTW, why not use smart pointers everywhere?

giuspen:
We could use smart pointers yes I'm not against just have been always used to clean in the destructors

PR #448, opened 2019-01-11
(Port to C++) popup menus

txe:
Added functions to generate different popup menus. 
Added showing node popup menu for the right button click on the tree node as an example. Although there is an issue, sometimes there is no event for button clicks on nodes, I have no idea why is so, tried signal_button_release_event and signal_button_press_event (behaves even worse).


giuspen:
Good stuff, thanks for your work. I've been very busy between family and work since 2019 started but I'll try to get back on track asap. As of the first thing I plan to work on is having the links work to improve navigation, then maybe the searches for the same reason.

PR #455, opened 2019-01-23
(Port to C++) node_add, node_duplicate

txe:
- Added `CtActions` to keep action functions. Also, there is `CtAction`, so the naming might be a bit confusing.
- Added actions: `node_add` and `node_duplicate`. The property dialog, its controls, and adding a node into the treestore are working, but some functions are still stubs. And when the node gets focus, sig fault occurs. I suppose, the problem is with uninitialized rTextBuffer.
- Added ct_dialogs where I will try to keep dialogs from support.py.
- I tried another approach in naming string and containers helper functions, like std::trim, str::join, str::split, str::endswith, vec::remove and etc. I think, this way it is easier to remember what to type. What do you think? Also, I keep the previous duplicate functions just in case.

BTW, can `ct_doc_rw.h`, `ct_xml_rw.cc`, `ct_sqlite3_rw` be renamed? It's just quite hard to look for them in IDE project tree while they are in totally different places. Maybe `ct_doc_rw.h`, `ct_doc_xml_rw.cc`, `ct_doc_sqlite3_rw.cc` or another way will be better?

I'll be out of town for the weekend, so I can start the next task next Monday.

### DIFF
--- a/future/Makefile.am
+++ b/future/Makefile.am
@@ -56,7 +56,9 @@ COMMON_SOURCES = \
 	src/ct/ct_p7za_iface.cc \
 	src/ct/icons.gresource.cc \
 	src/ct/ct_menu.cc \
-	src/ct/ct_pref_dlg.cc
+	src/ct/ct_pref_dlg.cc \
+	src/ct/ct_actions_tree.cc \
+	src/ct/ct_dialogs.cc
 
 cherrytree_SOURCES = \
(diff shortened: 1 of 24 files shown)


giuspen:
I'm happy of the utils renamings, just I would not keep duplicates.
Ok to rename ct_doc_*
Cheers :)

txe:
Ok, thanks!

PR #460, opened 2019-02-01
(Port to C++) node_date, node up/down/left/right

txe:
(ref to keep progress: #444)

- Added mentioned in the title actions.
- The hotkeys for these actions works ok, but there are still some issues that I try to fix later. 
- I implemented `widget_set_colors`, but it doesn't work as should. If you have time, can you look at it?

### DIFF
--- a/future/src/ct/ct_actions.h
+++ b/future/src/ct/ct_actions.h
@@ -13,8 +13,10 @@ private:
     CtTreeStore* _ctTreestore;
 
 private:
-    bool is_there_selected_node_or_error();
-    void _node_add(bool duplicate, bool add_child);
+    bool          is_there_selected_node_or_error();
+    void          _node_add(bool duplicate, bool add_child);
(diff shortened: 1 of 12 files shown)


giuspen:
@txe the widgets color is tricky moving to gtk3, it won't work unless you do it via stylesheets - see https://stackoverflow.com/questions/8952679/gtkmm-3-button-background-color-change
I would not worry though, I would put this low priority.
I spent some time this week trying to have the codebox taking key presses / text changes but I failed, I'll try again asap, my priority now is ensure that you can edit the anxhored widgets (codebox and table) and subsequently I would ensure the data can be written down to all document formats

giuspen:
Also @txe I can give you direct commit access to the cherrytree repo if that would ease your contributions

txe:
Thanks @giuspen it's good to know about stylesheets and I'm totally fine with pull requests, at least I hope the code will be reviewed and I won't  screw up too much.

PR #469, opened 2019-02-16
(Port to C++) view menu actions

txe:
(ref to keep progress: #444)

- Ported actions for view menu. For some reason, gtk doesn't return full-screen state, so `fullscreen_toggle` works noncompletely.
- Added missing copyrights


giuspen:
Cheers @txe if you would like to add yourself amongst the authors that would surely be fair

txe:
OK, thanks

PR #504, opened 2019-05-11
Styles revamp

gonzalezjo:
Fixes three issues in styles.css. 

The first bug is simple. Browsers get upset and issue a mixed-content warning when you import an insecure (http://) resource from a secure (https://) page. 

The second bug fixed is that every headings inserts newlines that do *not* exist in the original document. I sometimes use different headers on the same line in Cherrytree and, while the program itself renders them fine, the browser does not. Styling headings with     `display: inline;` fixes the issue.

The last and final bug that fixed is that whitespace does not render properly. If you have multiple levels of indentation in your notes and export to an HTML file, the indentation won't render. Applying the `white-space: pre-wrap;` style fixes the issue.


giuspen:
Cheers ;)

PR #512, opened 2019-06-08
(Port to C++, edit menu) Insert/Edit Link, insert timestamp, horizontal line, some bugfixes. 

txe:
Hi @giuspen, I again start working on the project, although slower this time.

I've made some bugfixes:

- Loading the config file throws an exception when a group is missing in the file
- CtXmlWrite::rich_txt_serialize toggles the case incorrectly. I made str::swapcase like swapcase in Python.

I ported from Edit menu:
- Insert/Edit Link
- Insert timestamp
- Insert horizontal line
- Changing the case

For changing the case I used some functionality from CtXmlWrite like rich_txt_serialize. As you are working on writing to xml document, could you add functions for writing pixbuf/table/codebox to xml as well?

(ref to keep progress: #444)


giuspen:
Hi @txe, super! I'm struggling these days between one kid and the other, your help is oxigen :)

giuspen:
As you are working on writing to xml document, could you add functions for writing pixbuf/table/codebox to xml as well?
=> yes that is what I'm about to do next

PR #524, opened 2019-07-09
(Port to C++, edit menu) paste-clipboard, duplicate row, move row up and down

txe:
(ref to keep progress: #444)
1. Add -lstdc++fs to work with std::filesystem
2. Add paste rich text/codebox/table/image/uri for the main viewer
3. More actions: text_row_selection_duplicate,  text_row_up, text_row_down

There is a small issue with tables. If copy & paste a table (through duplicate row, for example), then the pasted table takes several lines in the viewer, not one. But if click on the table and then click out of it, this fixes the table. Maybe it needs some redraw function after inserting, have you ever seen that bug?

As for my plans, I'm going to add signals and menus for anchored widgets, it should help with clipboard testing. And do you mind if I fix compiler warnings? It would be a separate pull request.

### DIFF
--- a/future/Makefile.am
+++ b/future/Makefile.am
@@ -293,11 +293,11 @@ libp7za_a_SOURCES = \
 
 cherrytree_LDADD = ${CHERRYTREE_LIBS} libp7za.a
 
-cherrytree_LDFLAGS = -Wl,--whole-archive $(top_srcdir)/libp7za.a -Wl,--no-whole-archive -lpthread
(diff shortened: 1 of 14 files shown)


giuspen:
Hi @txe, the gtkmm3 implementation of the tables is different from the old pygtk2 implementation since the old was a lists tore and this is a grid of textbuffers. I made this to pave the way for future enhancements of the tables while the list store implementation was limiting the improvements. So everything about that is new, not seen before.
Compiler warnings fixes are super welcome 😉 Cheers!

PR #560, opened 2019-09-15
update makefile to use flag -std=c++17

manscrober:
ct_actions.h and ct_actions_format.cc both use std::optional and include the optional header. This is a c++17 way of implementing it. Before it has to be std::experimental::optional from experimental/optional. Since I believe this was intended for c++17 I would say tell gcc to actually use c++17


giuspen:
It is already specified, look for AX_CXX_COMPILE_STDCXX_17 in configure.ac

manscrober:
apparently that's not enough. editing the cherrytree_cflags in the actual makefile does help though

manscrober:
@giuspen the addition is missing a \ in the end of the last line before my addition. I don't wnat to open another PR just for a type mistake but I did want to say something about it.

PR #604, opened 2020-01-05
Implement Command Palette Feature

ksmukta:
**Preview**: cherrytree-command-pallete
**Keybinding**: Ctrl+Shift+P
Currently all the menu items are shown in the command palette. User can search the palette by simply focusing and then typing in palette dialog window.
The search used is sub sequence search on lower cased strings.
**Note**: Ctrl+Shift+P was earlier used for PrintPageSetup but now its used for Command Pallete


giuspen:
The sub sequence search does not work for me (linux mint cinnamon latest), the search works only if you start typing from the very first letter but worse than the default search function.

ksmukta:
Could you please share some GIF to show how it does not work, Becuase for me it works fine(on Ubuntu 18.04).
command_palette_subsequence

giuspen:
Just try and type "pal", I expect "Command Palette" below to be focused, which it does not

txe:
For me, "tohtml" works on Ubuntu, later I can take a look on Arch.

txe:
The issue is with `is_subsequence`, which just checks individual letters , not as the whole text. I suggest to use something like this:
```python
data = data.tolower().replace(" ", "")
query = query.tolower().replace(" ", "")
return data.find(query) != -1
```

ksmukta:
Yes, need to update the search function.

ksmukta:
#622 
Fixed on this.
Thanks @txe for the code. Nice Idea.

ksmukta:
Currently the strings shown in command palette are using "dn"(description) from menudict. Can you help me how to use "sd"(short description). If I try to use "sd", the keys come with underscores like this: 

**New _Instance**, instead of **New Instance**

giuspen:
Those underscores are to generate so called accelerators, to quickly access the menu items using only the keyboard. There is a method to enable these accelerators, I'll see if I can find it browsing from my phone and get back to you

giuspen:
https://developer.gnome.org/pygtk/stable/class-gtkmenuitem.html I may have remembered wrong, try the set_use_underscore

PR #608, opened 2020-01-21
Add snap

hellsworth:
Until the C++ version is ready for release into the wild, I'd like to add snap packaging for the python 2.7 version of cherrytree. If you're not familiar with snap, it's similar to flatpak in that it is a containerized application that bundles its dependencies (like python 2.7) in the package, for use only by the app. 

Once the snap metadata is included in the project, I can setup an automatic launchpad build that will upload the cherrytree snap to the snap store. This is a great platform to distribute your package to users. In the future, when the C++ version is ready for release, we can change the launchpad build to build that version instead and even have it build automatically on every new commit added to master, so that the snap users are being provided the newest changes via the store in the edge branch (I would manually promote whichever build to the stable branch).

Snaps can be installed on any linux distribution with snapd installed (available from the distro repositories) and then run from the usual desktop icon or from the command line.

To build:
* the developer will need `snapd` installed
* the developer will need to use snap to install snapcraft and multipass from the snap store: `$ sudo snap install snapcraft multipass`
* then from the project root directory: `$ snapcraft`
* The result will be something like `cherrytree_0.36.6+git<id>_amd64.snap`

To test the build:
* Install the locally-built snap: `$ sudo snap install cherrytree_0.36.6+git<id>_amd64.snap --dangerous`
* Run cherrytree either from the desktop icon or from the command line: `$ snap run cherrytree`

### DIFF
--- a/snap/local/patches/desktop-launch.patch
+++ b/snap/local/patches/desktop-launch.patch
@@ -1,0 +1,34 @@
+diff --git a/common/desktop-exports b/common/desktop-exports
+index 1319361..cb9cd5d 100644
+--- a/common/desktop-exports
++++ b/common/desktop-exports
+@@ -94,6 +94,7 @@ append_dir LD_LIBRARY_PATH $RUNTIME/usr/lib/$ARCH/libunity
+ 
+ # Pulseaudio export
+ append_dir LD_LIBRARY_PATH $RUNTIME/usr/lib/$ARCH/pulseaudio
++append_dir LD_LIBRARY_PATH $SNAP/usr/lib/alsa-lib
+ 
+ # EGL vendor files on glvnd enabled systems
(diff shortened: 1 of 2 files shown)


giuspen:
I had a go this morning but the name cherrytree was already taken even though no package name cherrytree exists in snap. I started a dispute but the package may have to be called differently unfortunately. BTW @hellsworth thanks for your work

hellsworth:
@giuspen actually the name is taken by me so we don't need to change the name of the package in the snapstore. The snap is currently awaiting review before it's first allowed into the store (sometimes they can be slow) so hopefully it will be available soon.

giuspen:
Ok @hellsworth perfect cheers

hellsworth:
@giuspen you can try out the snap now from the snapstore: https://snapcraft.io/cherrytree

`sudo snap install cherrytree`

PR #621, opened 2020-02-04
Ensure the CtConst::AVAILABLE_LANGS array is the correct size

andygeorge:
In [this commit](https://github.com/giuspen/cherrytree/commit/f8dc01ccc4713355148b7c793bea7639f435eca8), the `sv` language identifier was added (per localization [done here](https://github.com/giuspen/cherrytree/pull/610)), but the array size was not increased, which causes a build failure. This PR increases the size.


giuspen:
Thanks, I shall change this in future to not have the size hard coded or I'll end up doing the same mistake

Issue #228, opened 2017-02-18
EcmaScript6 Syntax Hightlighting


kolomu:
Hello, it would be great if you can implement ES6 highlighting or atleast remove some valid ES6 words like class, import, ... from javascript. (ES6 is a newer JS version) 

I tried searching for the syntax highlighting feature but couldn't find it. 
Maybe you can give me a hint where I have to look so I can try implementing stuff like this on my own.

Thanks in advance.

giuspen:
Hi, the library in use for the syntax highlighting is gtksourceview.
You can find the syntax at https://developer.gnome.org/gtksourceview/stable/lang-tutorial.html
CherryTree is already hosting some language definitions in additions to the default available in the library in  the subfolder language-specs: https://github.com/giuspen/cherrytree/tree/master/language-specs
At the moment I cannot implement the new language spec myself but I will include it if you will work on it.
Cheers.

Issue #92, opened 2016-01-17
Feature Request: Rust language syntax


flukejones:
Would love to see syntax highlighting for the Rust language added.

https://www.rust-lang.org/

giuspen:
OK added for next version

Luke-Nukem:
Excellent! Thank you very much :smile:

Issue #38, opened 2015-05-24
Scala syntax highlighting


mnn:
It would be nice to have a Scala support. I was a bit surprised to not see it in the list, because it's more popular than e.g. F#, VHDL, Prolog or Haskell (at least according to Tiobe index).

giuspen:
I added it now, taking it from gtksourceview-3.0/language-specs (note that cherrytree is running gtksourceview-2.0).

mnn:
Thank you, it's working nicely ^_^.

PS: Installing necessary libraries under Windows is quite a pain (Pip is broken, so separate installers for libraries are required to be used).

txe:
Closing, the lang file was added

PR #634, opened 2020-02-08
Updates for lang specs (#228, #92, #38)

txe:
Hi, here:
- updates for some lang specs, taken from https://gitlab.gnome.org/GNOME/gtksourceview/tree/master/data/language-specs. I hope they are OK.
- new javascript.lang as a request from #228 
- also, issue   #92 can be closed because fulfilled

### DIFF
--- a/language-specs/go.lang
+++ b/language-specs/go.lang
@@ -31,41 +31,61 @@
   </metadata>
 
   <styles>
-    <style id="comment"           name="Comment"             map-to="def:comment"/>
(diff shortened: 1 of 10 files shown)


txe:
#38 can be closed too.

giuspen:
Just a note, sometimes the latest specifications have tags unsupported from the old pygtksourceview2; this is visible in warnings in runtime when you visit a code node in that language. If this happens, it is quite straightforward to remove the offending tag from the language spec file; I'll try and remember and check myself each of the modified langs before the next release anyway

Issue #434, opened 2018-11-26
[request] date node in a parent node "journal"


jhibern:
Hi,
Could someone help me to adapt **node_date** in **core.py** to insert date nodes (F8) in a parent node named "journal" or whatever?

I'm not a developer (my skills in python stop at modifying now_day date format in the way I prefer...), and I'd like to separate notes and journal nodes to better organize the tree and facilitate export. Thank you for your help.

txe:
Hi,
I like your idea, try this code (don't forget to make a backup just in case)
```python
   def node_date(self, *args):
        """Insert Date Node in Tree"""
        top_node = "journal"
        curr_time = time.time()
        now_year = support.get_timestamp_str("%Y", curr_time)
        now_month = support.get_timestamp_str("%B", curr_time)
        now_day = support.get_timestamp_str(self.journal_day_format, curr_time)
        self.node_child_exist_or_create(None, top_node)
        self.node_child_exist_or_create(self.curr_tree_iter, now_year)
        self.node_child_exist_or_create(self.curr_tree_iter, now_month)
        self.node_child_exist_or_create(self.curr_tree_iter, now_day)
```

jhibern:
Hi, thank you for your help.
I kept the old way to get time, but used your code to insert nodes. With my date format preferences, my node_date() method is now:
```python
    def node_date(self, *args):
        """Insert Date Node in Tree"""
        top_node = "journal"
        now_year = time.strftime("%Y").decode(locale.getlocale()[1])
        now_month =time.strftime("%Y-%B").decode(locale.getlocale()[1])
        now_day = time.strftime("%A %d %B %Y").decode(locale.getlocale()[1])
        #print now_year, now_month, now_day
        self.node_child_exist_or_create(None, top_node)
        self.node_child_exist_or_create(self.curr_tree_iter, now_year)
        self.node_child_exist_or_create(self.curr_tree_iter, now_month)
        self.node_child_exist_or_create(self.curr_tree_iter, now_day)
```
I don't understand why all thoses if/else in the native code while your code is so simple, but for now it seems to work in the way I wanted.
Thank you again.

PR #637, opened 2020-02-09
Fix visited nodes history going to todays node using key binding

ksmukta:
**Ctrl+T** is mapped to **Insert Today's Node**

**Before Fix**: Going to today's node adds Year, Month and Day nodes to visited history.
before_fix_history

**After Fix**: Just add Day node to visited nodes history.
after_fix_history

### DIFF
--- a/modules/core.py
+++ b/modules/core.py
@@ -2980,6 +2980,7 @@ iter_end, exclude_iter_sel_end=True)
 
     def node_date(self, *args):
         """Insert Date Node in Tree"""
+        self.last_command = "node_date"
         curr_time = time.time()
         now_year = support.get_timestamp_str("%Y", curr_time)
         now_month = support.get_timestamp_str("%B", curr_time)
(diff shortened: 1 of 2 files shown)


txe:
Can I suggest how to improve? I'm busy for now, but in couple of hours I will write how it can look.

txe:
1. `node_date` can be simplified and that will make easier to add your changes , try code from #434 .
2. before creating nodes, made a copy of `self.visited_nodes_list`
3. optional: after creating `year` and `month` nodes, remove their ids from a **copy** of `self.visited_node_list`
4. restore `self.visited_node_list` and create the last node

I think, it should work

txe:
Nice gifs, btw! If only everyone could add them in their issues :)

ksmukta:
> Nice gifs, btw! If only everyone could add them in their issues :)

Lol, thanks for noticing, takes time but makes things crystal clear.

ksmukta:
> 1. `node_date` can be simplified and that will make easier to add changes , try code from #434 .

I have created a pull request for this #638

ksmukta:
Closing this, fix pushed in #638

PR #644, opened 2020-02-12
Importing KeepNote to your cherry project take a long time

### LINKED ISSUE #643: Importing KeepNote to your cherry project take a long time
hi 
first thank u for this great tool that save my idea, codes and every think i needed in my life.
last days i find a great resource "keepNote resource here : https://github.com/souhaiboudiouf/OSCPRepo/tree/master/KeepNotes/BookmarkList" that i needed for my security learning as cherry tree lover i want to import it to cherry tree (new instance) so the problem was "Time", first i try to import all the keepnote resource but cherry tree crash and takes 8 GB  of my RAM in backround.
after that i take folder by folder not all the project and even that it takes a lot of time.

i have Dell I5 4th Generation/ 12 Ram 
i hope it will be an improvement in new cherry tree releases.

txe:
Resolve #643
It's quite big changes in core but one of them fixes a serious issue. Sorry, for the long text.

 There are two problems:
1.  **Really slow inserting of new nodes**. The problem is with finding a spare id, the algorithm `node_id_get` has O(n^2), when there are 20 nodes - it is fast, when 200 - it takes a second, when 800 - hmm.
So, firstly, I made it O(n) by finding max_node_id without checking every id.
Secondly, if it's import, first node uses (max_node_id + 1), second gets +1 without checking the whole tree (only checking `nodes_to_rm_set` just in case)
As a result 800 nodes are inserted in less a second. I think it will also make difference for other imports.
2. **Eating gigabits of memory by creating a lot of xml elements**. One of KeepNote documents is huge (5mb) and has a lot of space symbols and tags which generate a lot of elements. Because mostly elements have the same attributes, I can use the same xml element other and other without creating a new one. I has to use accumulator because just adding text by += is unbelievably slow.

### DIFF
--- a/modules/core.py
+++ b/modules/core.py
@@ -1436,6 +1436,19 @@ iter_end, exclude_iter_sel_end=True)
         elif event.button == 3: self.ui.get_widget("/SysTrayMenu").popup(None, None, None, event.button, event.time)
         return False
 
+    def max_node_id(self, tree_iter=None):
+        """Returns the max node id"""
+        if tree_iter == None: tree_iter = self.treestore.get_iter_first()
+        max_node_id = 0
+        while tree_iter != None:
+            max_node_id = max(max_node_id, self.treestore[tree_iter][3])
+            child_iter = self.treestore.iter_children(tree_iter)
+            while child_iter != None:
+                max_node_id = max(max_node_id, self.max_node_id(child_iter))
+                child_iter = self.treestore.iter_next(child_iter)
+            tree_iter = self.treestore.iter_next(tree_iter)
+        return max_node_id
+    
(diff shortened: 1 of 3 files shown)


giuspen:
Be careful, the valid node IDs start from 1, value 0 has a special meaning

txe:
Hi @giuspen , I checked `node_id_get` for:
- creating a node and subnode
- importing different formats and from cherrytree document
- the same as above but with removing nodes before operations
- the same as above but with not empty node tree

Speaking of Keepnote import, I found a couple of issue (have no idea how I overlooked them) but now it's ok.

giuspen:
Hi @txe excellent thanks for your work

PR #648, opened 2020-02-15
Add Collapse and Expand functionality in node tree of exported HTML

ksmukta:
When exporting tree to HTML, by default all the nodes are collapsed,
and the collapsed nodes have a '+' button which can be clicked to expand the
node.
* Expand/Collapse nodes in HTML
* ExpandAll/CollapseAll buttons included

improve_tree_in_exported_html

### DIFF
--- a/glade/script.js
+++ b/glade/script.js
@@ -1,7 +1,36 @@
 function changeFrame(file_path){
-	console.log(file_path);
 	var iframe = document.getElementById("page_frame");
 	iframe.src = file_path;
+}
+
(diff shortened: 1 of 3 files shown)


giuspen:
Thanks for your work. The distance between tree rows is very big, would it be possible to have them at least half distanced? I would rather have them all expanded by default since anyway it is just a click to collapse all.

ksmukta:
Done check now.

giuspen:
Excellent thanks

PR #671, opened 2020-03-07
(Port to C++) Undo\Redo

txe:
(ref to keep progress: #444)

Changes compared to pygtk: 
- Buffers with simple text also use machine_state, I couldn't find API for build-in undo\redo for them. Don't think it's a big deal.
- Add more parameters in Codebox constructor, so no need for additional method calls to setup codebox
- Add code to clone and compare widgets
- Didn't implemented `objects_buffer_refresh`, looks like it is a big hack for some reason, maybe code will work without it?

### DIFF
--- a/future/Makefile.am
+++ b/future/Makefile.am
@@ -73,7 +73,8 @@ COMMON_SOURCES = \
 	src/ct/ct_imports.cc \
 	src/ct/ct_widgets.cc \
 	src/ct/ct_print.cc \
-	src/ct/ct_export.cc
+        src/ct/ct_export.cc \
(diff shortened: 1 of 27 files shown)


giuspen:
Great job!
- objects_buffer_refresh was necessary in Pygtk2 implementation but definitely no longer needed in the much more efficient implementation I'm Gtkmm3
- the undo/redo API for plain text is into Gsv:: but is not strictly necessary, it will depend how good we can make our undo/redo since the Pygtk2 version is not so good and the library one was better for the plain text

Issue #668, opened 2020-03-06
HTML output


CGizmos:
I'm going to have to revert to a previous version. 

Links tree in every page is ignored - the tree is only there for index and no page gets shown with index. If I link to a specific html page, the index disappears.

Styles2.css seems to be ignored. I try to change h1 color/style and there is no effect.

The expanding index is a nice little effect, but not if it really breaks things.

txe:
I'll take a look how to fix it.

txe:
Hi,
I've made some fixes which, I hope, will help you.
You can either wait for the next release (probably, the next weekend), or update files manually if you used a standard installer (not a snap). The latter is better because in case of any issue I will fix it on-the-fly.
Files to update:
- `usr/share/cherrytree/modules/exports.py`
- `usr/share/cherrytree/glade/style2.css`
- `usr/share/cherrytree/glade/script.js`
New files you can take from the github (modules and glade folders).

In both cases you will need to remove 'style2.css' and 'script.js` from `.config/cherrytree`

txe:
Closing, it was fixed

PR #672, opened 2020-03-08
Fixing html export

txe:
Based on #668 issue:
- Fixed not working 'Links Tree in Every Page' option
- Fixed css, made class names more sensible
- Fixed button size (tree +/-) under Firefox
- Replaced 100vh, 100vw by absolute position to removed overflow-x/y = hidden
- Removed not used `node_export_to_html2`

**Edit:**

- Fixed title: add `<br/>` after title, add css customization

### DIFF
--- a/glade/script.js
+++ b/glade/script.js
@@ -34,5 +34,5 @@ function collapseAllSubtrees(element){
 }
 
 window.onload = function(){ 
-	document.body.style.overflowY = 'hidden';
+
 }
--- a/glade/styles2.css
+++ b/glade/styles2.css
@@ -21,51 +21,54 @@ Colours:
 body {
     font-size: 13px;
     font-family: "Open Sans", sans-serif !important;
-    padding: 20x; 
     white-space: pre-wrap;
-    margin: 0;
-    padding: 0;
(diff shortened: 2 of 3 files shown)


giuspen:
Many thanks for your gold work, I will also do some testings and likely release a minor update next weekend

txe:
:+1: Thanks

txe:
Probably, file names should be updated like style3.css and script3.js

giuspen:
I was also thinking that after reading your other comment :) I can do that one of the next day

giuspen:
Hi @txe if I double click on an html page directly rather than going through the index.html and clicking on its links tree, the index will not be shown while I think it is exactly this that @CGizmos meant in #668 - it would be great if you can take a look at that

txe:
Sure, I'll take a look a bit later

txe:
Hi @giuspen ,
Do you think it will be sufficient if i redirect from an html page to index.html and open that html page in the frame there?

giuspen:
Hi @txe I think so, if a much easier solution to try that first

giuspen:
Thanks @txe looks very good to me

PR #675, opened 2020-03-10
fixing "RuntimeError: returned character can not be represented in 16-bit unicode"

### LINKED ISSUE #664: Search Crashing Causes Nothing To Save After Event...
Loved the program, it's great for highlighting, organizing your work etc... I say loved because...

There's a nice bug that happens when you do a "Find In All Nodes Search" (possibly in other search cases as well).

It crashes, and lands on a random node, not one with the word in it. (That's the only way you know it's crashed). It would be great to have an error message come up saying something then, because if you choose to continue to work on anymore nodes adding and saving you will get a great surprise when you exit. The error message you would have enjoyed to have before suddenly shows up making and saving to a log file. It's great it saves that, because I truly don't know what I'd do without this info.

Hopefully it's useful to you...

Traceback (most recent call last):
  File "core.pyc", line 3249, in find_in_all_nodes
  File "findreplace.pyc", line 475, in find_in_all_nodes
  File "findreplace.pyc", line 787, in parse_given_node_content
  File "findreplace.pyc", line 746, in parse_node_content_iter
RuntimeError: returned character can not be represented in 16-bit unicode

The best part is although it saves this gibberish, every other thing that you have typed and saved into the other nodes after your search error has actually not been saved at all, you get this nice second surprise when you open it up again in CherryTree.

txe:
Resolves #664 
**Tested**: on windows (didn't test on linux), fonts Sans/Segue UI (maybe not important), a random text with symbol U+1F5F9 (taken from [wiki](https://en.wikipedia.org/wiki/Check_mark))
**Issue**: `TreeIter.get_char` throws exception `RuntimeError: returned character can not be represented in 16-bit unicode` on a bad symbols.
**Fixing**: I added `try/except` to mitigate the problem in places that throw the exception while I worked with the prepared text.
Probably, the issue wouldn't repeat with Gtk3, because it has a newer codebase.

### DIFF
--- a/modules/findreplace.py
+++ b/modules/findreplace.py
@@ -742,35 +742,38 @@ class FindReplace:
 
     def parse_node_content_iter(self, tree_iter, text_buffer, pattern, forward, first_fromsel, all_matches, first_node):
         """Returns True if pattern was find, False otherwise"""
-        buff_start_iter = text_buffer.get_start_iter()
-        if buff_start_iter.get_char() != cons.CHAR_NEWLINE:
-            self.newline_trick = True
-            if not text_buffer.get_modified(): restore_modified = True
-            else: restore_modified = False
-            text_buffer.insert(buff_start_iter, cons.CHAR_NEWLINE)
-        else:
-            self.newline_trick = False
-            restore_modified = False
-        if (first_fromsel and first_node)\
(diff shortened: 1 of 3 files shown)


giuspen:
Thanks @txe

PR #679, opened 2020-03-11
problem with opening another note

### LINKED ISSUE #677: Problem with opening another note on Ubuntu 16.04 using Cherrytree 0.39.0
Problem doesn't occur with Cherrytree 0.38.11.
Installed using deb packages:
cherrytree_0.38.11-0_all.deb
cherrytree_0.39.0-0_all.deb

I have many notes, ex.NoteA.ctb, NoteB.ctb, NoteC.ctb
When I double click NoteA.ctb it opens.
When I double click NoteB.ctb when NoteA is opened, instead of NoteB new empty instance of Cherrytree is opened. So now I have 2 instances opened: NoteA and empty.
When I double click NoteC.ctb nothing happens.

In Cherrytree 0.38.11 all 3 notes are opened in 3 instances, as expected.

txe:
Resolved #677
File name was unicode and converting it to unicode threw an exception.

### DIFF
--- a/modules/core.py
+++ b/modules/core.py
@@ -1507,7 +1507,10 @@ iter_end, exclude_iter_sel_end=True)
         """Try to load a file if there are the conditions"""
         #print "file_startup_load '%s' ('%s', '%s')" % (open_with_file, self.file_name, self.file_dir)
         if open_with_file:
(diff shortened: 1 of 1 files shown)


giuspen:
Thanks @txe

PR #680, opened 2020-03-12
Show node path in tooltip of Last Visited Nodes

### LINKED ISSUE #674: Show Node path in tooltip of Last Visited Nodes
In my Node tree I have Nodes for several years and each has a child called "Events". Looks like this:

```
2018
|-- Events
2019
|-- Events
2020
|-- Events
```

On the left of the Node Name Header I see the "Last Visited Nodes". In my case they all display "Events". I cannot see which year they belong to.

If I hover my mouse over the "Last Visited Nodes", a tooltip is already displayed which shows the full name of the Node.

My suggestion: Please extend the tooltip with the path of the Node. In my case the extended tooltip would be `2019/Events`.

txe:
Resolved #674


giuspen:
Cool thanks @txe

PR #681, opened 2020-03-14
(Port to C++) exec_code

txe:
(ref to keep progress: #444)

- ported exec_code
- fixed `get_font_family`, `get_font_size_str`, they didn't work with two-word font families.


giuspen:
Super, thanx @txe

PR #683, opened 2020-03-18
redirect from an html page to index.html

txe:
Based on  #672 discussion


giuspen:
Thanks @txe I'll give you feedback as soon as I can test it

PR #690, opened 2020-03-29
fixes regex in "Replace in Nodes Names"

### LINKED ISSUE #689: Wrong "." processing in "Replace in Nodes Names" (RegExp mode)
In "Regular expression" mode:
metacharacter "." works correctly in the "Replace in Node Content" operation,
but it’s wrong - in the operation "Replace in Nodes Names": it is treated as a symbol ".", not as a metacharacter, and only "." is replaced. The remaining characters in nodes names remain unchanged.

txe:
Resolves #689
I tested a normal mode and regex in "Replace in Nodes Names", they should work OK.


giuspen:
Thanks @txe!

PR #695, opened 2020-03-29
(Port to C+) Fixing some todo comments

txe:
(ref to keep progress: #444)


giuspen:
Thanks @txe good stuff

Issue #688, opened 2020-03-29
Use normal quotes


Y4kuzi:
How can I use normal quotes(' and ") in CherryTree 0.39.1 for Windows?
Disabling Smart Quotes did not work, I also tried to change the config.cfg but that didn't work either.
I enjoy using this notekeeping app but the abnormal quotes are really annoying.

txe:
Hmm, just checked on Win10, portable 0.39.1, with an old config (and after removing one).
Turned off `Enable Smart Quotes Auto Replacement` works as should, stopping converting quotes. The code is pretty simple there, I cannot see how it can be broken.

Y4kuzi:
Same issue with cherrytree_0.39.1_win32_portable. Unpacked it, made an empty config.cfg in the bin directory, opened the app, went to preferences and unchecked Enable Smart Quotes Auto Replacement. It still uses the weird quotes.

txe:
Just to be clear.
You type one quote, then second and after that they are replaced by weird looking quotes. Or the first one is already weird looking? Do you use US International Keyboard (layout)?

Y4kuzi:
My keyboard layout is set as NLD, which is my language.
I tried switching to US, just to test, and now the quotes are normal.
However, quotes now appear "instantly" after I press them, which isn't what I want.
Normally after I press on a quote key, it will wait for further input before displaying the quote. This is how I want it and should not change.
Any way to make it work with non-US keyboard layouts?

txe:
Yeah, it looks like this issue and #427 are the same. I checked a few pygtk2 programs and they all have that problem when such layouts are used and gtk engine just inserts odd quotes.

But `Enable Smart Quotes Auto Replacement` gave me an idea that these symbols can just be replaced with the right ones. I'll make the fix soon.

txe:
BTW, could you add "wrong" quotes here to be sure which one to replace?

PR #697, opened 2020-03-30
workaround for dead_diaeresis, dead_acute on win32

### LINKED ISSUE #427: International characters 'apostrophe' and 'quotation mark' not behaving as expected when using United States-International keyboard layout (Windows 7)
United States-International keyboard layout assigns special functions to a few keys (that is actually the main reason for using this layout, as it allow to write different languages without switching layouts).
From [https://support.microsoft.com/en-au/help/306560/how-to-use-the-united-states-international-keyboard-layout-in-windows](https://support.microsoft.com/en-au/help/306560/how-to-use-the-united-states-international-keyboard-layout-in-windows):

> When you press the APOSTROPHE ( ' ) key, QUOTATION MARK ( " ) key, ACCENT GRAVE ( ` ) key, TILDE ( ~ ) key, or ACCENT CIRCUMFLEX,. also called the CARET key, ( ^ ) key, nothing is displayed on the screen until you press a second key:
>- If you press one of the letters designated as eligible to receive an accent mark, the accented version of the letter appears.
>- If you press the key of a character that is not eligible to receive an accent mark, two separate characters appear.
>- If you press the space bar, the symbol (apostrophe, quotation mark, accent grave, tilde, accent circumflex or caret) is displayed by itself.

In cherrytree, both apostrophe and quotation marks keys ignore part of the functions and behave like if they were the special case symbol --i.e, the APOSTROPHE ( ' ) key always displays an accent ( ´ ) after pressing a second key, this behavior is the expected with the eligible keys, but when pressing another character or the space bar you expect an apostrophe.

This makes impossible to write apostrophe and quotation marks without switching keyboard layout. Though not a big issue from the user side, it is really odd as I never had this issue with any other software. And it could be considered an accessibility issue taking into account multilingual users who expect to write these characters as they are used to.

txe:
Partly fixes #427, #688 by replacing wrong 'apostrophe' and 'quotation mark' by the right ones. 
It only works for the main text buffer and codeboxes, text inputs and tables still are affected by the bug.

### DIFF
--- a/modules/core.py
+++ b/modules/core.py
@@ -63,6 +63,7 @@ class CherryTree:
         self.filetype = ""
         self.user_active = True
         self.ctrl_down = False
+        self.dead_key = ""
         self.window = gtk.Window()
         self.window.set_title("CherryTree")
         self.window.set_default_size(963, 630)
--- a/modules/support.py
+++ b/modules/support.py
@@ -344,6 +344,20 @@ def on_sourceview_event_after_key_press(dad, text_view, event, syntax_highl):
     """Called after every gtk.gdk.KEY_PRESS on the SourceView"""
     text_buffer = text_view.get_buffer()
     keyname = gtk.gdk.keyval_name(event.keyval)
+
+    # fix wrong diaeresis on win32 Internation keyboard after pressing SPACE
+    if cons.IS_WIN_OS:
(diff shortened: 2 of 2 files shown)


giuspen:
Thanks @txe

Issue #305, opened 2017-12-05
Keyboard shortcuts for sorting do not work


mikeyww:
Assigning a keyboard shortcut to a tree-sorting function [fails](https://www.giuspen.com/topic/keyboard-shortcuts/).  The keyboard shortcut has no effect.

Mike

Key words: sort tree keyboard shortcut

PR #706, opened 2020-04-01
Fixed https://github.com/giuspen/cherrytree/issues/305

Mammoets:
fixed https://github.com/giuspen/cherrytree/issues/305

### DIFF
--- a/modules/menus.py
+++ b/modules/menus.py
@@ -748,6 +748,21 @@ UI_INFO = """
     </menu>
 
     <menu action='TreeMenu'>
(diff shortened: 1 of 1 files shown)


txe:
The fix looks legit.
'TreeMenu' was filled by `self.menu_populate_popup` which puts accelerators into `self.orphan_accel_group` instead of global scope.

giuspen:
Thanks!

Mammoets:
Np! :)
Thank you!

PR #707, opened 2020-04-01
keep changed text in cell after table lost focus

### LINKED ISSUE #700: A table's cell content is lost unless you explicitly press Enter
Hello,

When you're typing inside a cell of a table, the usual behavior in any table-editing or spreadsheet-editing software is that if you click outside, the content you had already typed stays. 

In CherryTree, if you were at the middle of typing in a cell, and click outside of it for any reason, the cell's contents are immediately lost. It only gets "commited" to the cell if you explicitly press Enter.

In my opinion, the content shouldn't be lost. Since this works against the way tables work in most other software, it's really frustrating. I lose my text constantly because of it. Using the Undo function doesn't help much either, because it tends to "undo" several more changes previous to editing the cell.

Thanks for considering it.

txe:
Fixes #700

### DIFF
--- a/modules/tablez.py
+++ b/modules/tablez.py
@@ -455,11 +455,18 @@ class TablesHandler:
                 return True
         return False
 
+    def on_table_cell_focus_out_event(self, widget, event, path, model, col_num):
+        if model[path][col_num] != widget.get_text():
+            if self.dad.is_curr_node_not_read_only_or_error():
+                model[path][col_num] = widget.get_text()
+                self.dad.update_window_save_needed("nbuf", True)
+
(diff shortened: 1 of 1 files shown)


giuspen:
Nice one!

gnzgo:
Thank you so much, @txe! It's very kind of you.

txe:
You are welcome!
FYI, fixing your other issues will be postponed, they are not that critical and the current development focuses on the C++ version.

gnzgo:
I completely understand, @txe. I'm very thankful at any rate, it's so nice of you to contribute. So quickly, too! Thanks again.

PR #709, opened 2020-04-02
(Port to C++) zoom_text, zoom_tree, node_go_back, node_go_forward

txe:
(ref to keep progress: #444)

P.S.: I removed `_setFontForSyntax` and related functions because there is no need for them, fonts are already put in style classes (like `.rich-text`, etc.). and textviews only need to change its class to change font.


giuspen:
You don't sleep my friend :)

txe:
Well, then no surprise I woke up only now )

PR #710, opened 2020-04-02
(Port to C++) some fixes; theme for tree and textview

txe:
(ref to keep progress: #444)

- removed static ctMainWin from clipboard, there is no need for it
- changing colors for tree panel is working
- changing colors for textview will have two mode: light (classic) and dark (cobalt). Custom colors won't work.

CSS in terms of customizing gtksourceview is very limited and colors have to be set by theme manager (as I did now). The issue with custom colors is that bg and fg are not enough because other colors (selection, current-line, etc.) should be also changed. Gnote, for example, has a theme editor for customizing, to me, it looks easy to do, but can be postpone until first release.

It looks like, the only part to do are tables and then pre-release? 🎵 💃💃💃

### DIFF
--- a/future/src/ct/ct_actions.h
+++ b/future/src/ct/ct_actions.h
@@ -286,6 +286,10 @@ private:
 
 public:
     // others actions
+    void link_cut();
+    void link_copy();
+    void link_dismiss();
+    void link_delete();
     void anchor_cut();
     void anchor_copy();
     void anchor_delete();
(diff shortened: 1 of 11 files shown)


giuspen:
Wow🙃 You don't want to have a look at the tables too? 🤣 If too much no worries I can start to work on that in the weekend. On the other hand if you can try that I will instead try and focus on testing and spotting what is left to be ported.

txe:
Sure, firstly, I backport last fixes, then take on the tables.

PR #714, opened 2020-04-05
(Port to C++) start working on tables

txe:
(ref to keep progress: #444)

- Fixed table appearance.
- Add 'Insert Table'
- Unified css class names.
- Redone the state machine. Turned out keeping copies of anchered widgets was not a good idea because they responded to events like mouse clicks. Now, states of widgets are kept in CtAnchoredWidgetState which doesn't interact with UI.

Going to work on table actions.


giuspen:
Thanks @txe

PR #727, opened 2020-04-12
(Port to C++) CtStorageControl and refactoring

txe:
(ref to keep progress: #444)

About CtStorageControl
- reading and writing goes through CtStorageControl, then to either sqlite or xml storage. I suppose, adding a new storage (for example, to an online storage as google drive) will be easier.
- a backup will be created only **once** and only when saving is successful. I don't see much reason to create backups on every saving.
- a current shortcoming: if there is an error, user will see a dialog about it with pretty much nondescriptive message. But it's ok for testing.

Some fixes:
- Previously, popup menus didn't show nested submenus.
- Table headers were written incorrectly in C++, for pygtk they need to be at the end.

The current drawback, as I see, is that we cannot pass passwords and paths with spaces into 7z.  It should be fixed by passing data directly to main2.


giuspen:
Hi @txe, what do you mean by creating a backup only once? A backup that is too old won't be of much use.

txe:
The condition for creating a backup is to open a document and save some changes.
The next backup will be after reopening that document.

giuspen:
Some people have the document open for days (no joking) with never restart even the OS. I understand that to make a new backup at every save may be too much but also once every time you open the document may be ineffective for many. Anyway no worries I may add myself a configurable time since last backup.

txe:
Ah, OK, thanks

PR #738, opened 2020-04-19
Add travis ci

steveno:
This add a travis-ci script for the new C version of cherrytree.

I didn't modify the README to include the badge (since it's not ready yet) and you'll need to setup travis for the repository since I can't do that from an MR.

You can see the results of the build [here](https://travis-ci.org/github/steveno/cherrytree/builds/676766508).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-language: c
+language: c++
 cache: apt
 os: linux
 dist: bionic
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,20 @@
+language: c
+cache: apt
+os: linux
+dist: bionic
+
+compiler:
+  - gcc
+
+before_install:
+  - >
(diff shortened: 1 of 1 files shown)


txe:
How to setup travis: [short page](https://docs.travis-ci.com/user/tutorial/)

.travis.yml
```diff
@@ -0,0 +1,20 @@
+language: c
```
txe, line 1: I think, it should be c++?

steveno: In this case I'm not sure it mattered, but I did update it just in case.

steveno:
The newest build post my last commit.

https://travis-ci.org/github/steveno/cherrytree/builds/677058519

txe:
Thanks, I'll commit this PR, but I don't know when @giuspen will setup travis. 

Also, If you wish, you can add support for windows.

PR #748, opened 2020-04-24
add gspell for travis

txe:


txe:
well, to spend 25 mins to fix it is rather unfortunate

PR #751, opened 2020-04-26
(Port to C++) add cmake support and fix linker issue (-whole-archive)

txe:
(ref to keep progress: #444)

Hi, I made some fixes to remove necessity of `whole-archive`, and tests look fine under release and debug.
BTW, when you start cherrytree with the last changes, press Ctrl Shift P :)


giuspen:
Cool! I'll try and get access to the PC in the evening, already had a fight with wife in the morning so committed and got back to parent duty. BTW I installed 20.04 with the cinnamon desktop and it is so perfect.

PR #759, opened 2020-04-28
(Port to C++) use constexpr in ct_const.h; some tests for misc util

txe:
(ref to keep progress: #444)


giuspen:
Hi @txe let me fix the broken build I'm on Makefile.am right now for the package

txe:
Thanks

PR #776, opened 2020-05-03
Correct the Fedora build instructions

funnelfiasco:
Specifically, add the `make` step to perform the build and remove some unnecessary directory changes.

### DIFF
--- a/future/README.md
+++ b/future/README.md
@@ -99,9 +99,9 @@ sudo make install
 Get cherrytree source, compile and run:
 ```sh
 git clone https://github.com/giuspen/cherrytree.git
-mkdir cherrytree/build
-cd cherrytree/build
-cmake ../future
+cd cherrytree/future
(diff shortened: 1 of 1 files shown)


txe:
It's quite bad idea, cmake recommendations explicitly say not to use source dir as bulid dir. And though it's possible, it shouldn't be in README.

funnelfiasco:
I'm not very experienced with cmake, so I wasn't aware of that. But the README is still missing the `make` step. The instructions as written don't produce a binary.

txe:
Ah, yes, you're right. Would you like to make a PR? :)

funnelfiasco:
Yep, testing it now to make sure I have the directories correct. I don't have other OSes handy to test on, but I assume all of the Linuxes will need the same edit.

txe:
Right

PR #731, opened 2020-04-15
add powershell execution support

Kolossi:
A pretty simple one really, just to make powershell temp files save as "*.ps1" and add the code execution command to run them.

PR #785, opened 2020-05-06
add powershell execution support in non-future version

Kolossi:
ok, so this is embarrassing, my previous PR https://github.com/giuspen/cherrytree/pull/731 was merged but :

a) it only fixed the "future" version not current
b) it had a typo in that as well!

So _this_ PR should add powershell execution support to the current version and also fix the error in the future version.


giuspen:
Thanks @Kolossi !

PR #805, opened 2020-05-15
Add formatting of text without selecting it

ForeverRainbow:
This implements the feature request in #794 to have text formatting applied to future text rather than requiring text selection.

### DIFF
--- a/future/src/ct/ct_actions.h
+++ b/future/src/ct/ct_actions.h
@@ -258,7 +258,11 @@ private:
     bool _links_entries_pre_dialog(const Glib::ustring& curr_link, CtDialogs::CtLinkEntry& link_entry);
     Glib::ustring _links_entries_post_dialog(CtDialogs::CtLinkEntry& link_entry);
     Glib::ustring _link_check_around_cursor();
-
+    
+    Glib::ustring _current_prop_name;
+    Glib::ustring _current_prop_val;
+    bool _is_formatting = false;
+  
 public:
     // format actions
     void apply_tag_latest();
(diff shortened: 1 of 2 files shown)


txe:
It's not going to work, there are issues with code as well as UX

1. static var will work only once for one window and never for others
2. quite odd place for signal handler
3. in editors (e.g. MS Office) when you move cursor, current formatting changes (and look at 4rd point)
4. buttons (bold, italic, etc) never change their visual state, user don't know what is going on
5. in editors, when you change style, it applies for the current word, now it's broken.

ForeverRainbow:
So just to be clear, what needs fixing is:

- Proper logic for adding reconnecting after the TextBuffer is changed for node switch (1 and 2)
- When you change formatting, current word formatting should also change (i.e pressing `i` means if the cursor is currently in a word the whole word goes italic)
- Buttons need to have a visual "selected" effect (Using it I heartily agree :p)

I am confused about what you mean for 3, the tag will be applied to everything within a word (so if the cursor is moved between two formatted letters then any new letters will have the same formatting). Actually using it this is an issue because due to 5 you cannot change the style of a word without touching the mouse and selecting, is that what you mean?

txe:
I meant, in your PR, you remember the last used style, and after that any new text will always have that style. In editors, when user begins moving the cursor, styles under the cursor influences the current style. E.g., the style is bold, move cursor on other unbold word, the current style will be unbold.

Also, editors allows to apply several styles simultaneously; with remembering only one style, that's not going to happen. I suppose, there is a bunch of other small details we usually don't notice because good UI hides it from us.

txe:
OK, actually there is no reason to talk about features not related to your PR, otherwise it will be overwhelming.
Just implementation should not break the previous behavior too much.

giuspen:
My 2 cents saying I totally agree with @txe and this is the moment to deliver Gtkmm3 with no regression against the Pygtk2, a new pull request for an additional feature will be accepted at this time only if absolutely perfect. If people want to help please rather create pull requests fixing bugs or porting what is still missing.

jul-debug:
I don't know if this is an accepted practice on github but I want to thank yall for having taken this into consideration.

All the best.

PR #809, opened 2020-05-17
Fix searching map based on pointer key when exporting to pdf

ForeverRainbow:
This fixes a bug I encountered on my future-text-formatting branch, where exporting to pdf fails because the key lookup for `curr_attributes` fails (since it is comparing pointers). Weirdly this does not seem to be an issue upstream, possibly something to do with the changes to tags in my branch?

I am not sure why this ends up not working since it _should_ be comparing identical pointers (both read from `CtConst::TAG_PROPERTIES`). Regardless, comparing strings is a more robust option here.

It also includes formatting changes to improve the readability (e.g `static_cast` instead of C style casts, `auto` when casting, etc) and one or two optimisations (e.g move instead of copy)

### DIFF
--- a/future/src/ct/ct_export2pdf.cc
+++ b/future/src/ct/ct_export2pdf.cc
@@ -22,6 +22,8 @@
 #include "ct_export2pdf.h"
 #include "ct_dialogs.h"
 
+#include <utility>
+
 void CtExport2Pdf::node_export_print(const Glib::ustring& pdf_filepath, CtTreeIter tree_iter, const CtExportOptions& options, int sel_start, int sel_end)
 {
     std::vector<Glib::ustring> pango_slots;
@@ -49,7 +51,7 @@ void CtExport2Pdf::node_and_subnodes_export_print(const Glib::ustring& pdf_filep
     std::vector<Glib::ustring> tree_pango_slots;
     std::list<CtAnchoredWidget*> tree_widgets;
     Glib::ustring text_font = _pCtMainWin->get_ct_config()->codeFont;
-    _nodes_all_export_print_iter(tree_iter, options, tree_pango_slots, tree_widgets, text_font);
+    _nodes_all_export_print_iter(std::move(tree_iter), options, tree_pango_slots, tree_widgets, text_font);
 
     _pCtMainWin->get_ct_print().print_text(_pCtMainWin, pdf_filepath, tree_pango_slots, text_font, _pCtMainWin->get_ct_config()->codeFont,
                                            tree_widgets, _pCtMainWin->get_text_view().get_allocation().get_width());
@@ -106,7 +108,7 @@ void CtExport2Pdf::_nodes_all_export_print_iter(CtTreeIter tree_iter, const CtEx
         }
     }
     tree_widgets.insert(std::end(tree_widgets), std::begin(node_widgets), std::end(node_widgets));
-    for (auto iter: tree_iter->children())
+    for (auto& iter: tree_iter->children())
         _nodes_all_export_print_iter(_pCtMainWin->get_tree_store().to_ct_tree_iter(iter), options, tree_pango_slots, tree_widgets, text_font);
 }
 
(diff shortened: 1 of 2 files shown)


future/src/ct/ct_export2pdf.cc
```diff
@@ -49,7 +51,7 @@ void CtExport2Pdf::node_and_subnodes_export_print(const Glib::ustring& pdf_filep
     std::vector<Glib::ustring> tree_pango_slots;
     std::list<CtAnchoredWidget*> tree_widgets;
     Glib::ustring text_font = _pCtMainWin->get_ct_config()->codeFont;
-    _nodes_all_export_print_iter(tree_iter, options, tree_pango_slots, tree_widgets, text_font);
+    _nodes_all_export_print_iter(std::move(tree_iter), options, tree_pango_slots, tree_widgets, text_font);
```
txe, line 54: why?

future/src/ct/ct_export2pdf.cc
```diff
@@ -146,9 +148,9 @@ void CtPrint::print_text(CtMainWin* pCtMainWin, const Glib::ustring& pdf_filepat
     // convert to proxy widgets
     for (auto& widget: widgets)
     {
-        if (CtImage* image = dynamic_cast<CtImage*>(widget))            _widgets.push_back(std::shared_ptr<CtPrintImageProxy>(new CtPrintImageProxy(image)));
-        else if (CtTable* table = dynamic_cast<CtTable*>(widget))       _widgets.push_back(std::shared_ptr<CtPrintTableProxy>(new CtPrintTableProxy(table, 1, table->get_table_matrix().size())));
-        else if (CtCodebox* codebox = dynamic_cast<CtCodebox*>(widget)) _widgets.push_back(std::shared_ptr<CtPrintCodeboxProxy>(new CtPrintCodeboxProxy(codebox)));
+        if (auto* image = dynamic_cast<CtImage*>(widget))            _widgets.push_back(std::shared_ptr<CtPrintImageProxy>(new CtPrintImageProxy(image)));
+        else if (auto* table = dynamic_cast<CtTable*>(widget))       _widgets.push_back(std::shared_ptr<CtPrintTableProxy>(new CtPrintTableProxy(table, 1, table->get_table_matrix().size())));
+        else if (auto* codebox = dynamic_cast<CtCodebox*>(widget)) _widgets.push_back(std::shared_ptr<CtPrintCodeboxProxy>(new CtPrintCodeboxProxy(codebox)));
```
txe, line 153: When a block of few ifs, code after ifs are aligned

future/src/ct/ct_export2pdf.cc
```diff
@@ -325,7 +327,7 @@ void CtPrint::_on_draw_page_text(const Glib::RefPtr<Gtk::PrintContext>& context,
 {
     // layout num, line num
     std::pair<int, int> start_line_num = page_nr == 0 ? std::make_pair(0, 0) : print_data->page_breaks[page_nr - 1];
-    std::pair<int, int> end_line_num = page_nr < (int)print_data->page_breaks.size() ? print_data->page_breaks[page_nr] : std::make_pair((int)print_data->layout.size()-1, print_data->layout_num_lines.back());
+    std::pair<int, int> end_line_num = page_nr < static_cast<int>(print_data->page_breaks.size()) ? print_data->page_breaks[page_nr] : std::make_pair(static_cast<int>(print_data->layout.size()-1), print_data->layout_num_lines.back());
```
txe, line 330: that is why I hate unsigned int. Are you sure it will be ok when layout.size() == 0?
std::make_pair(static_cast<int>(print_data->layout.size()-1)

future/src/ct/ct_export2pdf.cc
```diff
@@ -364,9 +366,9 @@ void CtPrint::_on_draw_page_text(const Glib::RefPtr<Gtk::PrintContext>& context,
                 return;
         }
         // pixbuf or table or codebox
-        if (i < (int)print_data->layout.size() - 1) // the latest element is supposed to be text
+        if (i < static_cast<int>(print_data->layout.size() - 1)) // the latest element is supposed to be text
```
txe, line 369: I'd prefer
```static_cast<int>(print_data->layout.size()) -1```

future/src/ct/ct_export2pdf.cc
ForeverRainbow: `tree_iter` is passed by value, without the move it will be copied twice

txe: Well, I got it, but it is not optimization and does not improve readability:
- this place is called only once
- it's a single std::move in the whole project
- there are at least two more calls of _nodes_all_export_print_iter where std::move cannot be used

txe:
Thanks to finding this bug, it was introduced recently, when CtConst variables were moved from static to constexpr.
Static pointers were the same because they were in the same compiled unit (ct_const.o)
constexpr pointers, I suppose, don't have that quality.
This bug will affect other code too.

future/src/ct/ct_export2pdf.cc
txe: And compilers are smart enough to either add std::move themselves, when necessary, or ignore it

future/src/ct/ct_export2pdf.cc
TheGoldKnight23: Sorry my comment did not appear, yes this will cast properly (so `0 - 1` == `0` instead of wrapping)

PR #816, opened 2020-05-18
Implement HTML importing

ForeverRainbow:
This adds HTML importing to the C++ port.

Importing a node works as follows:
- Set-up the node and create its text buffer
- Run the `Html2Xml` parser on the input file's contents
- Run the parser output through `from_xml_string_to_buffer` and into the new node's text buffer
- Add the new node to the tree

Importing a directory simply iterates the files in the directory and runs the single file import function for each one with a `html` or `htm` extension.

I have also modified the HTML parser to parse indented lists as well as preventing it from pulling in the contents of some tags (e.g <script>).


txe:
Hi, thanks
Only some cosmetic changes are needed.

future/src/ct/ct_actions_import.cc
```diff
@@ -0,0 +1,92 @@
...
+ * MA 02110-1301, USA.
+ */
+
+#include "ct_actions.h"
+#include "ct_clipboard.h"
+#include "ct_imports.h"
+
+
+// Ask the user what file to import for the node
+void CtActions::import_node_from_html_file() noexcept {
```
txe, line 28: Functions have { on a new line

future/src/ct/ct_actions_import.cc
```diff
@@ -0,0 +1,92 @@
...
+    if (!_is_there_selected_node_or_error()) return;
+    nodeData.isBold = false;
+    nodeData.customIconId = 0;
+    nodeData.syntax = CtConst::RICH_TEXT_ID;
+    nodeData.isRO = false;
+    nodeData.name = filepath.stem();
+    
+    nodeData.rTextBuffer = _pCtMainWin->get_new_text_buffer();
+
+
```
txe, line 55: No need for second blank line (here and in other places)

future/src/ct/ct_actions_import.cc
```diff
@@ -0,0 +1,92 @@
...
+    CtHtml2Xml parser(_pCtMainWin);
+
+    try {
+
```
txe, line 59: no need for blank line after {

giuspen:
Thanks!

PR #823, opened 2020-05-22
Implement plaintext importing

ForeverRainbow:
This implements plaintext importing for the C++ port

It works basically the same as HTML importing except the text in the file is copied directly to the TextBuffer of the node instead of being parsed. So it works as follows
- Set-up the node and create its buffer
- Read the data out of the file and into a buffer
- Copy the buffer to the TextBuffer of the node
- Add the node to the tree

In order to properly detect text files I have also added a mimetype searching function to `CtMiscUtils`, it simply checks whether a certain substring appears in the mimetype for a file.

This also contains a bugfix (added 9 characters to the end of a line), which fixes the build error on windows because of the lack of an implicit conversion operator for `std::filesystem::path` to `std::string`


giuspen:
Thanks @ForeverRainbow good stuff! Could you add a unit test for CtMiscUtil::mime_type_contains() to tests/tests_misc_utils.cpp versioning a couple of lightweight files into tests/data please?

ForeverRainbow:
I have added a unit test, it generates its own dummy files in the directory its run and then cleans them up afterwards

giuspen:
I would rather have static dummy files so that can be opened from other text editors and verify their result for the encoding but if you cannot do that no worries I will

giuspen:
See https://github.com/giuspen/cherrytree/blob/master/future/tests/tests_tmp_n_p7zip.cpp and in particular const std::string unitTestsDataDir{Glib::build_filename(_CMAKE_SOURCE_DIR, "tests", "data")};

future/tests/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ set(CT_TEST_FILES
     tests_misc_utils.cpp
     tests_tmp_n_p7zip.cpp
     tests_types.cpp
+    test_consts.h
```
giuspen, line 11: The headers are not to be listed

future/tests/tests_misc_utils.cpp
```diff
@@ -311,6 +313,17 @@ TEST(MiscUtilsGroup, get_cherrytree_localedir)
...
 }
 
+TEST(MiscUtilsGroup, mime__type_contains) {
+    
+    
+
+    CHECK(CtMiscUtil::mime_type_contains(unitTestsDataDir+"/txt_test.txt", "text/"));
+    CHECK(CtMiscUtil::mime_type_contains(unitTestsDataDir+"/html_test.html", "text/"));
+    CHECK(CtMiscUtil::mime_type_contains(unitTestsDataDir+"/html_test.html", "html"));
+    CHECK(CtMiscUtil::mime_type_contains(unitTestsDataDir+"/cpp_test.cpp", "text/"));
```
giuspen, line 323: Can you also add a tiny binary and expecting not to find text mime?

giuspen:
Thanks!

PR #825, opened 2020-05-22
Implement cherrytree file importing 

ForeverRainbow:
This implements cherrytree file importing for the C++ port

Importing a node works as follows
- Read the node out of the external CT file
- Read the full buffer for the node out of the external CT file  
- Add node to the current tree in the appropriate position

(This is done for all the nodes in the file to be imported)

This also contains additions of foreign key's for `node_id` on most tables in the database


future/src/ct/ct_storage_control.cc
```diff
@@ -28,7 +28,17 @@
 #include <glib/gstdio.h>
 
 
-static CtStorageEntity* get_entity_by_type(CtMainWin* pCtMainWin, CtDocType file_type)
+Glib::ustring CtStorageControl::_check_and_unpack(const Glib::ustring& path) {
```
giuspen, line 31: I know it's not consistent at the moment but all future code (and to be gradually fixed in existing code) std::string is to be used for filepaths, not Glib::ustring

future/src/ct/ct_storage_control.cc
```diff
@@ -28,7 +28,17 @@
...
+    // unpack file if need
+    Glib::ustring password;
```
giuspen, line 33: also the password is just a string of plain bytes to be passed to p7zip so std::string is the right container

future/src/ct/ct_storage_control.h
```diff
@@ -58,8 +58,31 @@ class CtStorageControl
     void pending_new_db_node(gint64 node_id);
     void pending_rm_db_nodes(const std::vector<gint64>& node_ids);
     void pending_edit_db_bookmarks();
-private:
+    
+    constexpr CtStorageEntity* get_storage() const { return _storage; }
+    constexpr void set_storage(CtStorageEntity* storage) {
+        delete _storage;
```
giuspen, line 64: could we use a std::unique_ptr here and .reset()?

giuspen:
Hi @txe could you also have a look at these changes?

future/src/ct/ct_storage_control.h
```diff
@@ -29,7 +29,7 @@ class CtStorageControl
     static CtStorageControl* create_dummy_storage();
     static CtStorageControl* load_from(CtMainWin* pCtMainWin, const Glib::ustring& file_path, Glib::ustring& error);
     static CtStorageControl* save_as(CtMainWin* pCtMainWin, const Glib::ustring& file_path, const Glib::ustring& password, Glib::ustring& error);
-
+    static CtStorageEntity* get_entity_by_type(CtMainWin* pCtMainWin, CtDocType file_type);
```
txe, line 32: Should be removed or at least not public

future/src/ct/ct_storage_control.h
txe: I see no reason for get_storage and set_storage at all

txe:
@ForeverRainbow , can I know a reason why  you so actively edit/formatting existing source? It makes really hard to read changes and previous code is already testing

txe:
I don't think it's a good idea to turn on foreign keys, I would like to have full backwards compatibility for files for now.

ForeverRainbow:
Sorry, I added foreign keys because it makes it more explicit the relationship between the tables IMO. `get_entity_by_type` and `get_storage/set_storage` are there because I forgot to remove them after wrapping `import_nodes` in `add_nodes_from_storage`. Should I remove the foreign keys then? (It doesn't break backwards compatibility as far as I am aware)

giuspen:
The foreign key is fine if you can create a new SQLite document with your Gtkmm3 code then successfully read, edit and save with the Pygtk2 version. This is to be tested for such low level changes.

txe:
@ForeverRainbow,
it looks like you know how to code and I may look not grateful, but I am.

The reason for tension, the code was tested and you easily change it, especially storages. I tried really hard to make sure storages would work ok for all user cases and spend time to test for those user cases. And though your changes look small and harmless (like calling reopen_connection instead of sqlite3_open), I will have to do it again, because users can lose their data.

For best, new code are just new code without changing the existing code. And improvements and features are after release when we see that users don't have any issues with it.

ForeverRainbow:
Just tested it. I am able to create a cherrytree SQLite file on my fedora build of this and then open, edit, add nodes, and save it on my windows python version. Opening it on this version shows the changes (and checking the database, they are added correctly)

ForeverRainbow:
Ok, I apologise for that and if its any consolation I will also be testing it (since I am on fedora and this is the only way, apart from the snap, I can use this amazing application). The reason changed things was because I thought this was more of a testing stage and I wanted to make things easier later - using sqlite3_open_v2 for example will allow opening the database in read-only mode which will be helpful if multi-threading is to be used later

giuspen:
Looks good to me, ball on your court @txe

giuspen:
Thanks!

PR #830, opened 2020-05-23
Fix tables not saving with SQLite and remove enabling foreign keys

ForeverRainbow:
This one is on me

Turns out that widgets no longer save, because of course, the node is inserted before they are inserted and so the `has_*` field is never set :facepalm:, this was because I reordered the insertion order so as to abide by foreign key constraints

I have left the foreign keys in the schema as they are mostly for readability, but removed the pragma which enables them in every connection (and corrected the insertion order ofc).


giuspen:
Thanks @ForeverRainbow, I will prioritise the list of what is missing in the porting in a moment, the import from other proprietary note taking applications is for me one notch below the others

PR #832, opened 2020-05-23
Added check for ts_* fields when populating and saving the treestore

ForeverRainbow:
This fixes the schema mismatch error when importing CT SQLite files created before 49e7f9b for the C++ port

### DIFF
--- a/future/src/ct/ct_storage_sqlite.cc
+++ b/future/src/ct/ct_storage_sqlite.cc
@@ -837,27 +837,26 @@ void CtStorageSqlite::_check_db_tables()
     if (!_pDb) throw std::logic_error("CtStorageSqlite::_check_db_tables called without a valid _pDb object");
     
 
-    constexpr std::array<std::string_view, 2> needed_node_fields = {
-        "ts_creation INTEGER", "ts_lastsave INTEGER"
-    };
-    constexpr std::array<std::string_view, 3> needed_image_fields = {
-        "filename TEXT", "link TEXT", "time TEXT"
+    
+    const static std::vector<std::vector<std::string>> tables = {
+        {"node", "ts_creation", "INTEGER", "ts_lastsave", "INTEGER"}, {"image", "filename", "TEXT", "link", "TEXT", "time", "TEXT"}
     };
     
     try {
-        auto node_fields = get_table_field_names("node");
-        for (const auto& field : needed_node_fields) {
-            if (node_fields.find(field) == node_fields.end()) {
-                _exec_no_callback(fmt::format("ALTER TABLE node ADD COLUMN {}", field));
+        for (const auto& table : tables) {
+            auto& table_name = table[0];
+            auto node_fields = _get_table_field_names(table_name);
+            for (auto field = table.begin() + 1; field != table.end(); field += 2) {
+                if (node_fields.find(*field) == node_fields.end()) {
+                    auto sql = fmt::format("ALTER TABLE {} ADD COLUMN {} {}", table_name, *field, *(field + 1));
+                    _exec_no_callback(sql.c_str());
+                }
+                // Stop us going off the end
+                if ((field + 1) == table.end()) break;
             }
         }
 
-        auto image_fields = _get_table_field_names("image");
-        for (const auto& field : needed_image_fields) {
-            if (image_fields.find(field) == image_fields.end()) {
-                _exec_no_callback(fmt::format("ALTER TABLE image ADD COLUMN {}", field));
-            }
-        }
+        
 
     } catch(std::runtime_error& e) {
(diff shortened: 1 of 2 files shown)


future/src/ct/ct_storage_sqlite.cc
```diff
@@ -813,3 +815,37 @@ void CtStorageSqlite::_open_db(const std::string &path, bool read_only)
...
+
+    return fields;
+}
+
+void CtStorageSqlite::_check_db_tables(sqlite3* db) {
+    if (!db) throw std::logic_error("CtStorageSqlite::_check_db_tables passed invalid db object");
+    
+    auto fields = get_table_field_names(db, "node");
+
+    if (fields.find("ts_creation") == fields.end()) {
```
txe, line 840: take a look at db_tables_execute_check

future/src/ct/ct_storage_sqlite.cc
```diff
@@ -230,6 +231,7 @@ bool CtStorageSqlite::save_treestore(const Glib::ustring& file_path, const CtSto
         {
             _open_db(file_path, false);
 
+            _check_db_tables(_pDb);
```
txe, line 234: no tables, no need

future/src/ct/ct_storage_sqlite.cc
txe: to prev comment: there are other columns

future/src/ct/ct_storage_sqlite.cc
```diff
@@ -813,3 +815,37 @@ void CtStorageSqlite::_open_db(const std::string &path, bool read_only)
        // _exec_no_callback(foreign_keys_pragma);
     }
 }
+
+
+std::unordered_set<std::string> get_table_field_names(sqlite3* db, const std::string& table_name) {
```
txe, line 820: you can easily make it a private function

future/src/ct/ct_storage_sqlite.cc
```diff
@@ -813,3 +815,37 @@ void CtStorageSqlite::_open_db(const std::string &path, bool read_only)
...
+        // Name is the second column
+        std::string name = reinterpret_cast<const char*>(sqlite3_column_text(stmt, 1));
+        fields.emplace(std::move(name));
+    }
+
+    return fields;
+}
+
+void CtStorageSqlite::_check_db_tables(sqlite3* db) {
+    if (!db) throw std::logic_error("CtStorageSqlite::_check_db_tables passed invalid db object");
```
txe, line 836: no need to pass db, use _pDb, as a result not need to check it

future/src/ct/ct_storage_sqlite.cc
ForeverRainbow: Whoops, yeah put that in the wrong place, it does need to go in the other branch of the if statement though right?

ForeverRainbow: Wait nvm, if opening an old file `populate_treestore` will be called before anyway

future/src/ct/ct_storage_sqlite.h
```diff
@@ -57,6 +57,20 @@ class CtStorageSqlite : public CtStorageEntity
     Gtk::TreeIter       _node_from_db(guint node_id, Gtk::TreeIter parent_iter);
     CtTreeIter          _node_from_imported_db(gint64 node_id, CtTreeIter* parent_iter);
     
+    /**
+     * @brief Check that the database contains the required tables
+     * 
+     * @param db 
+     */
+    void                _check_db_tables(sqlite3* db);
```
txe, line 65: wrong signature

txe: in .cc its _check_db_tables()

ForeverRainbow:
This won't built, fixing it one sec

future/src/ct/ct_storage_sqlite.h
ForeverRainbow: Yep, I pushed it accidentally :facepalm: (meant to do `git diff`, history was `git push`)

future/src/ct/ct_storage_sqlite.cc
```diff
@@ -813,3 +814,53 @@ void CtStorageSqlite::_open_db(const std::string &path, bool read_only)
...
+        for (const auto& table : tables) {
+            auto& table_name = table[0];
+            auto node_fields = _get_table_field_names(table_name);
+            for (auto field = table.begin() + 1; field != table.end(); field += 2) {
+                if (node_fields.find(*field) == node_fields.end()) {
+                    auto sql = fmt::format("ALTER TABLE {} ADD COLUMN {} {}", table_name, *field, *(field + 1));
+                    _exec_no_callback(sql.c_str());
+                }
+                // Stop us going off the end
+                if ((field + 1) == table.end()) break;
```
txe, line 855: this condition will never be true

PR #838, opened 2020-05-25
Fix pointer arithmetic when creating lists

ForeverRainbow:
Found a similar one to this when doing the lists for HTML, possibly a change in `CtConfig` but anyway without wrapping it in a `Glib::ustring` (`std::string`'s char ctor is not utf8 aware AFAIK) it is performing pointer arithmetic before adding the text


txe:
Well, this bug is nasty, you can find more by looking up "[0] +"
Maybe it will be better to convert charsTodo (and others) to vector of std::string

ForeverRainbow:
Shall I try and find the rest and push it to this?

txe:
std::wchar_t is a different beast and shouldn't be used here because it has a fixed symbol size while UTF-8 (which is general for gtk) has variable symbol size.
std::string or Glib::ustring should be better

ForeverRainbow:
OK, should I convert them to `std::string` and push it?

txe:
Yes

ForeverRainbow:
Ok this is almost done I think, was a _little_ more complex than I though, still got a utf8 issue which I am fixing now

ForeverRainbow:
Ok this is working completely now I think

txe:
Ok, I'll take a look tomorrow

txe:
`ct_const.h ` doesn't actually have changes, so can be reverted to keep history simple.

future/src/ct/ct_list.cc
```diff
@@ -87,7 +87,7 @@ void CtList::list_handler(CtListType target_list_num_id)
                     new_par_offset = range.iter_end.get_offset() + 2;
                     end_offset += 2;
                     int bull_idx = (!list_info) ? 0 : (list_info.level % (int)_pCtMainWin->get_ct_config()->charsListbul.size());
-                    _curr_buffer->insert(range.iter_start, _pCtMainWin->get_ct_config()->charsListbul.substr((size_t)bull_idx, 1) + CtConst::CHAR_SPACE);
+                    _curr_buffer->insert(range.iter_start, _pCtMainWin->get_ct_config()->charsListbul.item().substr((size_t)bull_idx, 1) + CtConst::CHAR_SPACE);
```
txe, line 90: use [] instead  substr

future/src/ct/ct_list.cc
```diff
@@ -167,14 +167,15 @@ CtListInfo CtList::list_get_number_n_level(Gtk::TextIter iter_first_paragraph)
     int level = 0;
     while (iter_start) {
         auto ch = iter_start.get_char();
-        if (str::indexOf(_pCtMainWin->get_ct_config()->charsListbul, ch) != -1) {
+        auto& bulls_list = _pCtMainWin->get_ct_config()->charsListbul.item();
+        if (str::indexOf(bulls_list, ch) != -1) {
```
txe, line 171: I'd add indexOf into the class, to make class more useful, but it's up to you

future/src/ct/ct_splittable.h
```diff
@@ -0,0 +1,203 @@
...
+        return _internal_vec;
+    }
+    
+    CtSplittable(const CtSplittable&) = default;
+    CtSplittable(CtSplittable&&) noexcept = default;
+    CtSplittable(ITEM_T item, ITEM_T_SPLITTER item_splitter = &split_items<ITEM_T>) : _item_splitter(item_splitter), _item_cache(std::move(item)), _set_vec_cache(false) {}
+    
+    CtSplittable& operator=(CtSplittable other)
+    {
+        swap(*this, other);
```
txe, line 108: doesn't look great, = and swap have different meaning, also because of that you don't use const ref

txe:
That's all, after fixing I'll commit it

future/src/ct/ct_splittable.h
ForeverRainbow: This is the copy and swap idiom, [this so post](https://stackoverflow.com/a/3279550/12448530) explains it pretty well

txe: Ok, got it

future/src/ct/ct_list.cc
ForeverRainbow: I'll add a `find` specialisation for `char` to `CtStringSplittable`

ForeverRainbow: On seconds thoughts maybe not, just realised I would have to take `char` by const ref

txe:
Great, thanks! let's hope there is no more troubles with this pointer arithmetic

PR #853, opened 2020-05-27
Implement Zim importing

ForeverRainbow:
I think this is feature complete (apart from image importing), the parser can handle
- Header tags 1-5 (caps at 3 and sets 4-5 to 3)
- Bold text
- Italic text
- Strike-through text
- https and http links
- Lists (ordered, bullet, todo)
- Verbatim text
- Subscript
- Superscript
- Inter-page links (to inter-node links)

It handles recursive building and parsing of the Zim tree, mimicking its structure in the nodes it creates.

It does **not** handle multiple formatting at once, but that is not an issue because Zim cannot handle it either. It will generate excess `<rich_text></rich_text>` tags if formatted text appears consecutively but I do not think this is really an issue (if it is, it should not be hard to fix)


txe:
Hi, looks good.
Although I wonder why CtSplittable is here again, it's some git glitch?

About `CtActions::_add_node_with_data`, I suppose, this function is going to be used only for imported nodes. If so, I can suggest:
- remove everything after (and including) `if (node_state) {`. These functions are slow and actually don't make sense when a number of nodes is created. Call these functions once after importing.
- then naturally it needs other name, e.g.`_add_node_fast` or something like that

Are you creating buffers after all because links should be fixed, and they can be fixed only after nodes are created? If yes, can you add comments there.

ForeverRainbow:
I think I did add comments? If not oops and I shall fix that, `_add_node_with_data` is actually lifted straight from `_node_add_with_data` which I did not want to modify, but I needed it to return the `Gtk::TreeIter`, I will move those out of the function

ForeverRainbow:
Also do I not need to call `pending_new_db_node()` when creating the new node, overwise I think it will not save?

txe:
I tried to understand why buffers were created at the end and not together with nodes and it takes some time to figure out. Comments don't clarify that.

pending_new_db_node is needed, right.

code for ctsplittable was merged or just copy/paste? I'm asking just out of curios

ForeverRainbow:
CtSplittable was cherry-picked from the master branch

txe:
Thanks for comments.

I just read a bit about cherry picking and turned out it usually has some issues. Probably your branch can be merged just fine without that commit. You can revert that commit by (it will create one more commit, but it's ok)
- `git revert 6a0e65c`
- commit the reverted changes

And I hurried with 'removing all', `tree_store.update_node_aux_icon(node_iter);` in `_add_node_fast` is still needed. Sorry.

txe:
Thanks!

PR #876, opened 2020-06-04
command to duplicate node with subnodes

### LINKED ISSUE #318: Feature Request - Duplicate Node with SubNodes
Hi,

I'm currently trying to migrate from Keepnote which has a feature to copy and paste the whole node tree. 

It would be great if the same kind of functionality, which would allow for duplicating the node along with its subnodes could be added to cherrytree.

### LINKED ISSUE #360: Duplicating node doesn't copy children
I've just updated CherryTree and gone to duplicate a node, when I do so, I'm no longer copying any children and only copies the parent node with nothing ending up underneath it.

Please let me know any other further information I can provide, I can't share the document and was unsure what else would prove useful. This happens in the Windows CherryTree 0.38.4 build.

### LINKED ISSUE #501: Node/Subnode duplication, templating
Say we wanted to do this:

-> [Node] (as a category)
--> Template (sub-node + other sub-nodes)
---> Subnode [X]
---> Subnode [Xn] (both using the template)
-> [Node X] (duplicated from [Node] with the template only).
-> [Node Xn] (duplicated from [Node] with the template only).

How do we achieve this?

If templating was supported, we could then click "new X", name it and populate the available fields with ease, instead of export import .....

[Image](https://maikthulhu.github.io/img/onenote-layout.jpg)
This is what I'm referring to.

It would also be very nice to have a "dynamic" templates, for example, subnode-X has a property "Windows". It could then change the rest of the template for that child to follow Windows-related tasks/steps defined in the template, making pen-testing steps / documentation a lot easier. Template within a template (for a host with X VMs linux/windows/whatever).

You see where I'm going with this.

### LINKED ISSUE #506: [Feature request] Subnode capabilities
It'd be great if Sub Nodes could load, parse (and store) git .md files content from either an URL or a local clone.

### LINKED ISSUE #872: [Feature request] Duplicate node tree
I am duplicating some node trees and unless I am missing a trick I cannot duplicate the sub-nodes (the sub tree structure) under a selected node. Only the parent node is duplicated.

At Tree > Duplicate Node i suggest another option Duplicate Node and Sub Nodes (the full depth)..

txe:
Hi @giuspen, I added a new command to duplicate a node with subnodes.
For now, there is no proper image and title. Do you have any?

Resolves #872
Resolves #506
Resolves #501
Resolves #360
Resolves #318

### DIFF
--- a/future/src/ct/ct_actions.h
+++ b/future/src/ct/ct_actions.h
@@ -165,9 +165,10 @@ private:
     Gtk::TreeIter _add_node_quick(const Gtk::TreeIter& curr_iter, CtNodeData& node_data, bool is_child);
 public:
     // tree actions
-    void node_add()       { _node_add(false, false); }
-    void node_dublicate() { _node_add(true, false);  }
-    void node_child_add() { _node_add(false, true); }
+    void node_add()                { _node_add(false, false); }
+    void node_dublicate()          { _node_add(true, false);  }
+    void node_child_add()          { _node_add(false, true); }
+    void node_subnodes_dublicate();
     void node_edit();
     void node_inherit_syntax();
     void node_delete();
@@ -177,7 +178,7 @@ public:
     void node_down();
     void node_right();
     void node_left();
-    void node_change_father();
+    void node_change_father();    
     bool node_move(Gtk::TreeModel::Path src_path, Gtk::TreeModel::Path dest_path);
     void tree_sort_ascending();
     void tree_sort_descending();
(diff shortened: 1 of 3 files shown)


giuspen:
Hi @txe, good stuff! For the .SVG icon we have to create one using inkscape starting from existing ones, I can do that! Not sure what you mean with title

txe:
Good, thanks then I just submit this.

PR #879, opened 2020-06-05
Fix apostrophe spell check

AyubM:
Hi, I couldn't find the same file to edit in the future directory, so I hope making the change on master is ok.

The spellchecker flagged words with apostrophes as wrong, e.g. "didn't" would be highlighted. I Googled and found a patch that works, but it seems they never submitted it as a pull request so I'm going ahead and doing it here.

Implements https://gist.github.com/hcs64/993e77d4539f84ca33ca#file-spellcheck_apostrophe-patch


giuspen:
Thanks!

PR #880, opened 2020-06-05
Add spdlog logging

ForeverRainbow:
For this I have basically gone through the source and converted every `std::cout` into an info, warn, or debug message and turned every `std::cerr` into an error message. I have also added a handler for gtk log messages which sends them to a separate spdlog sink.

The log level is set to debug for both compile-time and runtime currently by default

This is linking to the header-only version of spdlog because I could not get it to work with the compiled version, [this](https://github.com/gabime/spdlog/issues/1366#issuecomment-567748773) comment suggest that it may be better to just use the header version if the sources are going to be included into `src`.


giuspen:
Thanks @ForeverRainbow I'll wait for @txe to approve this

future/src/ct/ct_clipboard.cc
```diff
@@ -640,7 +641,7 @@ void CtClipboard::_on_received_to_uri_list(const Gtk::SelectionData& selection_d
                 else
                 {
                     property_value = "";
-                    std::cout << "ERROR: discarded file uri " << file_path << std::endl;
+                    //spdlog::error("discarded file uri {}", file_path);
```
txe, line 644: why is it commented?

future/src/ct/ct_main_win.cc
```diff
@@ -341,7 +343,7 @@ const std::string CtMainWin::get_text_tag_name_exist_or_create(const std::string
         }
         if (not identified)
         {
-            std::cerr << "!! unsupported propertyName=" << propertyName << " propertyValue=" << propertyValue << std::endl;
+           // spdlog::error("!! unsupported propertyName={} propertyValue={}", propertyName, propertyValue);
```
txe, line 346: why is it commented?

txe:
It looks good, thanks!

ForeverRainbow:
Damn those should not have been commented sorry, it was when testing the Glib::ustring logging and I though I got them all

PR #888, opened 2020-06-07
download image when image is copied from browser

txe:
- download image when image is copied from browser
- made a function to download stuff from the internet, sorry for the name


txe:
Wow, cool number - 888 😄

giuspen:
Good stuff! I'm not far from finishing the tree info but got to pause for family duty, I'll try and complete later

PR #893, opened 2020-06-08
Add python3-urllib3 package and network plug. Fixes #887.

### LINKED ISSUE #887: [snap] copy image from browser
Hi @hellsworth , cherrytree uses `urllib2` python package for downloading images from the internet. For example, when users copy text with images from the browser and insert them into a note, instead of image links, cherrytree downloads images to paste them.

So, I see snap is missing `urllib2` python package, can it be added?
And does snap need network privileges or something like that?

hellsworth:


txe:
Thanks!

Issue #878, opened 2020-06-05
Implement `pandoc` support, rather than a bunch of on-off format translators.


DiagonalArg:
Discussion moved over from Issue #858 

[`Pandoc`](https://en.wikipedia.org/wiki/Pandoc) is a "general markup converter," that covers [dozens of formats](https://en.wikipedia.org/wiki/Pandoc#Supported_file_formats) and supports plugs for new formats, to be written in [Lua](https://en.wikipedia.org/wiki/Lua_(programming_language)).  The subject came up in the above issue, because of its extensive coverage of various markdown flavors, which can be mixed with LaTeX:

From the man page:

> Pandoc's enhanced version of Markdown includes syntax for footnotes, tables, flexible ordered lists, definition lists, fenced code blocks, superscripts and subscripts, strikeout, metadata blocks, automatic tables of contents, embedded LaTeX math, citations, and Markdown inside HTML block elements. (These enhancements ... can be disabled using the markdown_strict input or out‐put format.)

Pandoc runs on [both Windows and any unix-like system](https://pandoc.org/installing.html), where it is in most linux repositories and available both as a package on the [pandoc site](https://pandoc.org/), and in Homebrew.  Many projects supply interfaces between their project and pandoc. For example, [Sublime](https://packagecontrol.io/packages/Pandoc) and [Atom](https://atom.io/packages/pandoc-convert). There are many more, as this is a well known and heavily used tool. It's the swiss army knife of document conversion. 

> Pandoc is a free and open-source document converter, widely used as a writing tool (especially by scholars) and as a basis for publishing workflows. It was created by John MacFarlane, a philosophy professor at the University of California, Berkeley. (From Wikipedia.)

If a Cherrytree to/from translator for pandoc were developed, then that Lua script could be included with Cherrytree.  With pandoc as an optional dependency, it could become the one tool needed for anything from [Makdown + LaTeX to rtf & html to pdf & epub, etc. etc. etc](https://pandoc.org/).

Right now would be the time to do it, as it would save the work of writing a markdown translator, while at the same time offering a much more advanced markdown dialect including, for free, LaTeX.

txe:
I don't know, I would say no than yes

- Cherrytree supports a small number styles. 
- Markdown is a quite simple format. It can be easily implemented in our case. 
- Latex, rtf and others are used by user to create rather complicated and stylized texts. 

[...]

PR #903, opened 2020-06-11
Implement importing through pandoc

ForeverRainbow:
This implements some of #878, with the ability to import files by running the contents through Pandoc.

Currently it will only work if the pandoc executable is in the user's PATH. This is more or less an arbitrary restriction which just needs an option to be added to the preferences and used instead of the hard-coded `pandoc` but I think it is quite minor and maybe not needed.

It does work on windows using the win32 api, however I am getting compile + link times of 20+ minutes (just for ct_process) which I am not sure if is a problem with my computer or mingw being slow. 

Exporting will be next I think, but is going to be difficult to keep both cherrytree "native" formatting and whatever formatting pandoc applies. It may have to be done on a select text to export basis.

### DIFF
--- a/future/src/ct/CMakeLists.txt
+++ b/future/src/ct/CMakeLists.txt
@@ -60,6 +60,8 @@ set(CT_SHARED_FILES
     ct_text_parser.cc
     ct_md_parser.cc
     ct_parser.cc
+    ct_pandoc.cc
+    ct_process.cc
     )
 
 add_library(cherrytree_shared STATIC ${CT_SHARED_FILES})
--- a/future/src/ct/ct_actions.h
+++ b/future/src/ct/ct_actions.h
@@ -402,6 +402,7 @@ private:
     void _import_node_from_plaintext(const std::filesystem::path& filepath);
     void _import_nodes_from_zim_directory(const std::filesystem::path& filepath);
     void _import_node_from_md_file(const std::filesystem::path& filepath);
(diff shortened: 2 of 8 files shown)


txe:
I haven't known how you were going to use process, so couldn't give an advice.
Mostly platform-depending things are already implemented in glib, so you can just use https://developer.gnome.org/glib/stable/glib-Spawning-Processes.html

ct_pandoc files are really small, can you place them in ct_export2html

if ct_process will have less code as a result of using glib, or even the class can be replaced by one function, it can be put in ct_misc_utils.

ForeverRainbow:
I really wish I had known about that _before_ messing with win32 :p. OK so would it be better if I reimplement this using GSubProcess? I mean it currently works so :woman_shrugging: 

Even using GSubProcess I would prefer to keep `ct_process` as an interface for it just for the things we need, I can put it in `ct_misc_utils` but that means including the entirety of `ct_misc_utils` just for spawning a child process.

txe:
OK

ForeverRainbow:
So should I change it to use GSubProcess?

giuspen:
@ForeverRainbow I would prefer if you can use glib as much as possible for cross platform utilities

ForeverRainbow:
Ok, I'll close this for now then and rewrite it

giuspen:
Thanks!

PR #905, opened 2020-06-11
Implement pandoc importing support

ForeverRainbow:
This is a revised version of #903 which uses GSubprocess for management of subprocesses (i.e pandoc)

It also includes the functionality to import a directory from pandoc, however it has not been added to the UI since I think there will need to be some configuration screen with accepted formats becuase right now it will just try and send every file it can find through pandoc which obviously doesn't turn out well.

It also fixes an issue I discovered while testing, where trying to import a binary file using pandoc (e.g an odt) wouldn't work.

I have also noticed an issue where tables are inserted into the currently selected node rather than the node they are meant to be imported into. I assume this is to do with how the html parser handles tables


txe:
g_find_program_in_path  - to find the program

I'm pretty sure, code in action_imports can be written much simpler, to import from pandoc file one function calls another and that calls another and the same couple of time more. 

I don't like how `add_nodes_with_directory_hierarchy` is made, it has templates, callbacks, options. calls other template functions. Looks universal, but depth is hardcoded. It can be much simpler

pandoc code also can be in export2html.cc

g_subprocess_new returns process pointer, should it be clean up?
stdout_stream  shouldn't be closed?

ForeverRainbow:
Whoops, yes it should thanks for spotting that!

The stdout_stream does not need to be closed, it is done automatically when the GSubprocess is destroyed. 

`add_nodes_with_directory_hierarchy` is that way so that it can be used for other importers when importing directories, it is recursive because it needs to be to handle subdirectories and depth is not needed for now but will be if a maximum depth restriction is imposed. The templates are just so that any valid functor can be used. I am not sure how it can be simplified while still maintaining a node hierarchy based on the filesystem.

The pandoc code should be in export2html, I missed that sorry

txe:
some glib functions return Glib::RefPtr and some not. With auto and unknown functions, it's really hard to read what is going. If it's possible, add types to variables

ForeverRainbow:
I remember now why I didn't delete `process`, `GSubprocess` is opaque and doesn't seem to offer any destructor, I am assuming that `g_subprocess_wait` will destroy it once done?

txe:
Yeah, I couldn't find out about it too

txe:
Thanks, don't worry about unifying import functions, I was going to do about it to simplify unit testing.
And thanks for the issue with tables, hope I won't forget to fix it.

ForeverRainbow:
I'll open an issue for the tables thing

PR #908, opened 2020-06-13
Set architecture to any

steev:
This lets us build the package on arm64, armhf, armel and i386 instead
of just amd64.


steev:
I've been running with this change for a while locally, to build the c++ version of Cherry Tree on Kali for my arm64 machines.

giuspen:
Ok thanks

PR #909, opened 2020-06-13
Implement Table of Contents inserting for C++

ForeverRainbow:
This implements Table of Contents inserting for the C++ port.

I think this matches the behaviour of the python version 100%, it uses the same anchor structure for nodes (`h<level>-<num>`) and the same character set for the list bullets.

There is a bug which I noticed which causes a segfault if an anchor is removed and then another node is selected. This is a general issue with anchors in the C++ version I think since I get the same bug if I use "insert anchor".

### DIFF
--- a/future/src/ct/ct_actions.h
+++ b/future/src/ct/ct_actions.h
@@ -33,6 +33,7 @@
 #include <filesystem>
 
 class CtMainWin;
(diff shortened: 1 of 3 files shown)


giuspen:
Thanks!

PR #914, opened 2020-06-14
Add CSV handling to CtTable in the C++ port

ForeverRainbow:
Supports excel style csv with comma field delims and double quotes for strings.

The reason it for the ugly code in `from_csv` where it reads the data in chunks is because it needs to be able to handle newlines in cells

### DIFF
--- a/future/src/ct/ct_actions_edit.cc
+++ b/future/src/ct/ct_actions_edit.cc
@@ -95,6 +95,7 @@ void CtActions::table_handle()
         while ((int)rows.size() < _pCtMainWin->get_ct_config()->tableRows)
             rows.push_back(empty_row);
     }
+    CtTable* pCtTable = nullptr;
     if (res == 2) {
         CtDialogs::file_select_args args(_pCtMainWin);
         args.curr_folder = _pCtMainWin->get_ct_config()->pickDirCsv;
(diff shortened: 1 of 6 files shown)


giuspen:
The Windows build is behaving strangely after I made it build in Debug so no panic if it keeps failing I'll switch back to Release

txe:
`row_to_csv` is not good, text can have " or \n or maybe something else, any of that breaks csv format.

ForeverRainbow:
The newline is not an issue since it gets escaped into a string anyway, the `"` is an issue

txe:
the same thing with reading, do you support escaping " (like \")? because it's quite often used.

Also, it's not easy to test, I'd prefer to have vectors of string vectors to read/write. It will be double job to convert them to matrix, but it's easy to test.

ForeverRainbow:
Working on escaping now, I can split `from_csv` into public methods which return `CtTableRow`, `CtTableCell` and `CtTableMatrix` which can be used for testing?

ForeverRainbow:
Also what do you mean by 
> vectors of strings to read/write
?

txe:
all classes you named create buffers and do a lot of work which currently hard to fake in tests. And idea is to test how csv reading\writing works because it's can have issues.

As I see, there will be one function which reads a file and creates vector<vector<string>>. The result is used to create matrix. You can wrap it up in one function call  `from_csv`

Same thing with converting matrix to csv.

ForeverRainbow:
Ah I see, so create a matrix of the cell contents before wrapping them into `CtTableCell`, I agree thats much cleaner :p

txe:
can you put CtCSV in ct_misc, so ct_table will be cleaner? Then I accept this PR

ForeverRainbow:
@txe Makes more sense in `ct_parser`?

txe:
No, I would put it in ct_parser if it was based on ctparser. It just a couple of functions to read/write csv file.

txe:
Thanks

PR #917, opened 2020-06-15
Feature improved markdown

ForeverRainbow:
This is a much improved version of the markdown parser. 

It adds
- Codeboxes
- Tables
- Headers using `==` and `----`
- Images 
- Monospace
- Horizontal spacers
- Bold and italic at the same time

And fixes several issues with conflicting symbols.

This also includes functions for image, codebox, and table xml serialising under the `CtXML` namespace. I have not changed the existing functions in `CtHtml2Xml` or the respective widgets but they can now be changed to use the needed function instead of duplication.

Note that widgets (such as codeboxes, tables, and images) are still inserted into the active node when import is selected. Which I am guessing has something to do with `from_xml_to_buffer`.

### DIFF
--- a/future/src/ct/ct_imports.cc
+++ b/future/src/ct/ct_imports.cc
@@ -44,6 +44,31 @@ xmlpp::Element* codebox_to_xml(xmlpp::Element* parent, const Glib::ustring& just
     p_codebox_node->set_attribute("show_line_numbers", std::to_string(show_line_numbers));
     return p_codebox_node;
 }
+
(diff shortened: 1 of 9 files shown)


txe:
`get_uri_type` doesn't look reliable, because url can have `file://` or `d:\\\\`, but I couldn't find out a proper function to replace it.

About widgets appearing on the current node, I know where the issue is, I'll fix it.

ForeverRainbow:
@txe `file://` should be added to the check for local files, if it is `d:\\` (windows d drive I assume you mean?) then the `std::filesystem::exists` check should get it, its imperfect though I agree

PR #920, opened 2020-06-15
Use Pandoc when exporting to HTML

ForeverRainbow:
This adds Pandoc usage to the HTML parser. If the syntax format is supported (currently `markdown` and `latex`) then it will be run through pandoc (with a warning if pandoc could not be found). This works for codeboxes and nodes with syntax highlighting.

I have also added a switch in `CtConfig` as `usePandoc` which currently is always true but can be used to disable pandoc exporting in the future (i.e if the user wants unformatted text for some reason)

Ref #878

### DIFF
--- a/future/src/ct/ct_export2html.cc
+++ b/future/src/ct/ct_export2html.cc
@@ -351,7 +351,8 @@ Glib::ustring CtExport2Html::_html_get_from_code_buffer(const Glib::RefPtr<Gsv::
     code_buffer->ensure_highlight(curr_iter, code_buffer->end());
     if (_pCtMainWin->get_ct_config()->usePandoc && CtPandoc::supports_syntax(syntax_highlighting)) {
         if (CtPandoc::has_pandoc()) {
-            Glib::ustring      input_txt(curr_iter, code_buffer->end());
+            Gtk::TextIter end_iter = sel_end >= 0 ? code_buffer->get_iter_at_offset(sel_end) : code_buffer->end();
+            Glib::ustring      input_txt(curr_iter, end_iter);
             std::istringstream ibuff(input_txt);
             std::ostringstream html_out;
             CtPandoc::to_html(ibuff, html_out, syntax_highlighting);
@@ -678,7 +679,7 @@ bool has_pandoc()
 }
 
 
-void to_html(std::istream& input, std::ostream& output, std::string from_format /* = "markdown" */)
+void to_html(std::istream& input, std::ostream& output, std::string from_format)
 {
     auto process = pandoc_process();
     process->append_arg("--from");
--- a/future/src/ct/ct_export2html.cc
+++ b/future/src/ct/ct_export2html.cc
@@ -123,7 +123,7 @@ void CtExport2Html::node_export_to_html(CtTreeIter tree_iter, const CtExportOpti
         }
     }
     else
-        html_text += _html_get_from_code_buffer(tree_iter.get_node_text_buffer(), sel_start, sel_end);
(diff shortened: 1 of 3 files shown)


txe:
@giuspen what do you think?

future/src/ct/ct_export2html.cc
```diff
@@ -345,11 +345,23 @@ Glib::ustring CtExport2Html::_get_table_html(CtTable* table)
...
 // Get rich text from syntax highlighted code node
-Glib::ustring CtExport2Html::_html_get_from_code_buffer(Glib::RefPtr<Gsv::Buffer> code_buffer, int sel_start, int sel_end)
+Glib::ustring CtExport2Html::_html_get_from_code_buffer(const Glib::RefPtr<Gsv::Buffer>& code_buffer, int sel_start, int sel_end, const std::string& syntax_highlighting)
 {
     Gtk::TextIter curr_iter = sel_start >= 0 ? code_buffer->get_iter_at_offset(sel_start) : code_buffer->begin();
     code_buffer->ensure_highlight(curr_iter, code_buffer->end());
-    Glib::ustring html_text = "";
+    if (_pCtMainWin->get_ct_config()->usePandoc && CtPandoc::supports_syntax(syntax_highlighting)) {
+        if (CtPandoc::has_pandoc()) {
+            Glib::ustring      input_txt(curr_iter, code_buffer->end());
```
txe, line 354: not a `code_buffer->end()` but `sel_end`
function is used to get html from selected text

future/src/ct/ct_export2html.cc
```diff
@@ -666,9 +678,11 @@ bool has_pandoc()
 }
 
 
-void to_html(std::istream& input, std::ostream& output) 
+void to_html(std::istream& input, std::ostream& output, std::string from_format /* = "markdown" */)
```
txe, line 681: it's not necessary
but it may be unclear that `to_html` implicitly uses `markdown` style. maybe it would be better if format will be explicit

ForeverRainbow: Pandoc defaults to markdown thats why it's needed, I didn't make it explicit since the expected behaviour of pandoc if you do not specify a format is markdown but I can see how that could be unclear xD

txe:
Thanks, the only issue I see, how user will know of such improvement? or it can even puzzle user when the same document will give different results on different machines.

giuspen:
I'm not sure what are the differences between the html produced with or without pandoc for the code nodes and codeboxes. Depending on this we should decide whether it is the case to have a configuration option for this in the preferences dialog and what the default value should be.

txe:
@giuspen , in this case, pandoc is used for nodes with markdown/latex syntax, so the output will have rich format text in html. Without pandoc, the output will be plain text in html. So there's a difference

giuspen:
@txe I understand now thanks, I also think the user should be notified in a non invasive way that pandoc was looked for but not found, maybe the status bar could be a compromise or a dialog with the possibility to check 'don't show anymore'

ForeverRainbow:
@giuspen Currently it logs a warning that the format could have been formatted using pandoc but pandoc could not be found, it then continues and uses the plain text instead. Since `CtExport2Html` has a `CtMainWin` that could very easily have a popup added.

btw I was thinking, using spdlog a custom logger could be created which creates the appropriate dialogue box for the user, only issue would be making sure it always had a valid pointer to `CtMainWin`

txe:
It's quite bad idea that spdlog will do some side job with dialogs and will know something about which main win is current

giuspen:
yep the logger is only for us or maybe for an expert user but for the normal user I would push a message to the status bar or show a dialog. Actually since GTK3 there are additional quite good in app notifications that we may start to use.

PR #924, opened 2020-06-16
Improve HTML output

ForeverRainbow:
This improves the styling of the output HTML, mostly it only uses `style3.css` but there is also some html injection using `script3.js`.

Changes:
- Tree panel can be shown/hidden using a button between it and the `page_frame` iframe, mousing over the button when closed will open it
- Background is now `#EEEEEE` for everything
- Backgrounds for widgets (i.e tables) is now `rgb(197, 197, 197)` (`#FFC1C1C1`)
- Links on the tree expand to the full width of the row and the row lights up (`rgba(197, 197, 197, 0.7)`) on mouse over instead of the link being underlined
- The `+/-` buttons on the tree no longer have borders and are vertically centred in the row
- The `index` text at the top of the tree panel has been moved up and made `h1`
- Buttons no longer have borders or backgrounds and instead their text is lightened when moused over
- Headers for node titles have been centred, moved up, and given a background so that they at least somewhat resemble the node headers in cherrytree
- Codeboxes are now `inline-block`, meaning they will resize to fit their contents instead of being the width of the page (with a minimum width of `30%`)
- Codeboxes and tables now have a black border instead of grey
- Tables now have solid borders instead of disjointed lines

I realise that users can do all of this themselves by changing `style3.css` and `script3.js` in the config folder but I think these are better "defaults" as it were. They will of course not take affect unless the user replaces their existing config files

I have also changed the C++ version when exporting tables to correct the issue of headers being all the far left cells instead of the first row and squashed table cells due to empty contents

### DIFF
--- a/future/data/script3.js
+++ b/future/data/script3.js
@@ -1,3 +1,7 @@
+right_arrow_heavy = '&#129090;'
+left_arrow_heavy = '&#129088;'
+
+
 function changeFrame(file_path){
 	var iframe = document.getElementById("page_frame");
 	iframe.src = file_path;
@@ -5,7 +9,7 @@ function changeFrame(file_path){
 
 function toggleSubTree(element){
 	var nextSibling = element.parentElement.nextElementSibling;
-	if(nextSibling.tagName == 'UL'){
+	if(nextSibling.tagName === 'UL'){
 		nextSibling.classList.toggle('hide');
 		if(element.textContent === '+'){
 			element.textContent = '-'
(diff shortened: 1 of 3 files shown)


txe:
It looks ok, but did you try the output in different browsers?

ForeverRainbow:
Tested it in chrome, chromium and firefox. As far as I can tell its identical across all of them

txe:
@giuspen , I think files again have to change numbers 3 -> 4
Maybe we need to stop copying these files into user config

giuspen:
@txe the issue is that some users want to be able to customize the stylesheet

ForeverRainbow:
Maybe it could be changed to use the priority convention of numbers at the end, so `style-3.css`, would be included before `style-4.css` (and thereby overridden)

PR #925, opened 2020-06-16
Implement custom link actions

ForeverRainbow:
This implements custom link actions for the C++ port.

It invokes the external application using `std::system`, GSubprocess is a bit heavy for this I think but it does have the issue that output of the child muddies the output of cherrytree 

It also adds a `path` class to `CtFileSystem` which is basically a version of `std::filesystem::path` with only the bits that are currently needed. I added this because I needed a way to format paths correctly across platforms and didn't want to write `get_proper_platform_path` every time, but it also has `operator/` overloaded for `g_build_filename`.

### DIFF
--- a/future/src/ct/ct_actions_others.cc
+++ b/future/src/ct/ct_actions_others.cc
@@ -30,6 +30,7 @@
 #include <fstream>
 #include <cstdlib>
 #include "ct_logging.h"
+#include <spdlog/fmt/bundled/printf.h>
 
 // Cut Link
 void CtActions::link_cut()
@@ -158,7 +159,7 @@ void CtActions::embfile_open()
 
     spdlog::debug("embfile_open {}", static_cast<std::string>(filepath));
 
-    CtFileSystem::external_filepath_open(filepath, false);
+    CtFileSystem::external_filepath_open(filepath, false, _pCtMainWin->get_ct_config());
     _embfiles_opened[filepath] = CtFileSystem::getmtime(filepath);
 
     if (not _embfiles_timeout_connection)
@@ -264,7 +265,13 @@ void CtActions::link_clicked(const Glib::ustring& tag_property_value, bool from_
          Glib::ustring clean_weblink = str::replace(vec[1], "amp;", "");
          if (_pCtMainWin->get_ct_config()->weblinkCustomOn)
          {
-             // todo: subprocess.call(self.weblink_custom_action[1] % clean_weblink, shell=True)
+             std::string cmd = fmt::sprintf(_pCtMainWin->get_ct_config()->weblinkCustomAct, clean_weblink);
+             int retr = std::system(cmd.c_str());
+             if (retr == -1) {
+                 // Internal std::system error
+                 spdlog::error("Error while executing: '{}'; Message: {}", cmd, std::strerror(errno));
+                 return;
+             }
          }
          else g_app_info_launch_default_for_uri(clean_weblink.c_str(), nullptr, nullptr); // todo: ?
      }
@@ -278,7 +285,7 @@ void CtActions::link_clicked(const Glib::ustring& tag_property_value, bool from_
          }
          if (from_wheel)
              filepath = Glib::path_get_dirname(CtFileSystem::abspath(filepath));
-         CtFileSystem::external_filepath_open(filepath, true);
+         CtFileSystem::external_filepath_open(filepath, true, _pCtMainWin->get_ct_config());
      }
      else if (vec[0] == CtConst::LINK_TYPE_FOLD) // link to folder
      {
@@ -290,7 +297,7 @@ void CtActions::link_clicked(const Glib::ustring& tag_property_value, bool from_
          }
          if (from_wheel)
              folderpath = Glib::path_get_dirname(CtFileSystem::abspath(folderpath));
-         CtFileSystem::external_folderpath_open(folderpath);
+         CtFileSystem::external_folderpath_open(folderpath, _pCtMainWin->get_ct_config());
      }
      else if (vec[0] == CtConst::LINK_TYPE_NODE) // link to a tree node
      {
--- a/future/src/ct/ct_actions_others.cc
+++ b/future/src/ct/ct_actions_others.cc
@@ -32,6 +32,8 @@
 #include "ct_logging.h"
 #include <spdlog/fmt/bundled/printf.h>
 
+namespace fs = CtFileSystem;
+
 // Cut Link
 void CtActions::link_cut()
 {
@@ -159,7 +161,7 @@ void CtActions::embfile_open()
 
     spdlog::debug("embfile_open {}", static_cast<std::string>(filepath));
 
-    CtFileSystem::external_filepath_open(filepath, false, _pCtMainWin->get_ct_config());
+    CtFileSystem::external_filepath_open(filepath.c_str(), false, _pCtMainWin->get_ct_config());
     _embfiles_opened[filepath] = CtFileSystem::getmtime(filepath);
 
     if (not _embfiles_timeout_connection)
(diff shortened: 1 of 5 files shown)


giuspen:
I have few comments please wait before merge

future/src/ct/ct_misc_utils.cc
```diff
@@ -992,61 +994,51 @@ std::string CtFileSystem::get_file_stem(const std::string& path)
...
-        subprocess.call(self.filelink_custom_action[1] % filepath, shell=True)
-    else:
-        if cons.IS_WIN_OS:
-            try: os.startfile(filepath)
-            except:
-                if open_fold_if_no_app_error: os.startfile(os.path.dirname(filepath))
-        else: subprocess.call(config.LINK_CUSTOM_ACTION_DEFAULT_FILE % re.escape(filepath), shell=True)
-        */
+bool CtFileSystem::exists(const CtFileSystem::path& filepath) {
+    return g_file_test(filepath.c_str(), G_FILE_TEST_EXISTS);
```
giuspen, line 998: please use Glib::file_test(filepath, Glib::FILE_TEST_EXISTS)

txe:
I cannot review or answer today, the internet connection is really bad now

future/src/ct/ct_misc_utils.h
```diff
@@ -439,4 +443,46 @@ std::string get_cherrytree_lang_filepath();
...
 
+/**
+ * @class path
+ * @brief An object representing a filepath
+ */
+class path {
+    using path_type = std::string;
+public:
+    path(path_type path) : _path(std::move(path)) {}
+    path(const Glib::ustring& path) : _path(path.begin(), path.end()) {}
```
giuspen, line 454: Glib::ustring is not to be used for filepaths, please use std::string instead

future/src/ct/ct_misc_utils.h
```diff
@@ -439,4 +443,46 @@ std::string get_cherrytree_lang_filepath();
...
+    path(const char* path) : _path(path) {}
+    
+    ~path() = default;
+    
+    void append(const path& other) {
+        _path = g_build_filename(_path.c_str(), other.c_str(), nullptr);
```
giuspen, line 460: please use Glib::build_filename(_path, other);

ForeverRainbow:
I have removed the ability to use `Glib::ustring` however there are some cases where I have constructed `CtFileSystem::path` from `Glib::ustring` using `c_str()` as otherwise a whole bunch of code has to be changed

future/src/ct/ct_actions_others.cc
```diff
@@ -158,7 +161,7 @@ void CtActions::embfile_open()
 
     spdlog::debug("embfile_open {}", static_cast<std::string>(filepath));
 
-    CtFileSystem::external_filepath_open(filepath, false);
+    CtFileSystem::external_filepath_open(filepath.c_str(), false, _pCtMainWin->get_ct_config());
```
giuspen, line 164: Why .c_str()? Maybe .string()?

future/src/ct/ct_actions_others.cc
```diff
@@ -264,33 +267,39 @@ void CtActions::link_clicked(const Glib::ustring& tag_property_value, bool from_
...
      else if (vec[0] == CtConst::LINK_TYPE_FILE) // link to file
      {
-         Glib::ustring filepath = CtExport2Html::_link_process_filepath(vec[1]);
-         if (not Glib::file_test(filepath, Glib::FILE_TEST_IS_REGULAR))
+         fs::path filepath = CtExport2Html::_link_process_filepath(vec[1]).c_str();
+         if (not Glib::file_test(filepath.string(), Glib::FILE_TEST_IS_REGULAR))
          {
-             CtDialogs::error_dialog(str::format(_("The File Link '%s' is Not Valid"), std::string(filepath)), *_pCtMainWin);
+             CtDialogs::error_dialog(fmt::format("The File Link '{}' is Not Valid", filepath), *_pCtMainWin);
              return;
```
giuspen, line 286: You have removed a translatable string here _("translatable string"), is it intentional?

future/src/ct/ct_actions_others.cc
```diff
@@ -264,33 +267,39 @@ void CtActions::link_clicked(const Glib::ustring& tag_property_value, bool from_
...
      }
      else if (vec[0] == CtConst::LINK_TYPE_FOLD) // link to folder
      {
-         Glib::ustring folderpath = CtExport2Html::_link_process_folderpath(vec[1]);
-         if (not Glib::file_test(folderpath, Glib::FILE_TEST_IS_DIR))
+         fs::path folderpath = CtExport2Html::_link_process_folderpath(vec[1]).c_str();
+         if (not fs::is_directory(folderpath))
          {
-             CtDialogs::error_dialog(str::format(_("The Folder Link '%s' is Not Valid"), std::string(folderpath)), *_pCtMainWin);
+             CtDialogs::error_dialog(fmt::format("The Folder Link '{}' is Not Valid", folderpath), *_pCtMainWin);
```
giuspen, line 297: You have removed a translatable string here _("translatable string"), is it intentional?

future/src/ct/ct_misc_utils.cc
```diff
@@ -992,61 +998,51 @@ std::string CtFileSystem::get_file_stem(const std::string& path)
...
-        subprocess.call(self.filelink_custom_action[1] % filepath, shell=True)
-    else:
-        if cons.IS_WIN_OS:
-            try: os.startfile(filepath)
-            except:
-                if open_fold_if_no_app_error: os.startfile(os.path.dirname(filepath))
-        else: subprocess.call(config.LINK_CUSTOM_ACTION_DEFAULT_FILE % re.escape(filepath), shell=True)
-        */
+bool CtFileSystem::exists(const CtFileSystem::path& filepath) {
+    return Glib::file_test(filepath.c_str(), Glib::FILE_TEST_EXISTS);
```
giuspen, line 1002: Again why .c_str(), maybe .string()?

future/src/ct/ct_misc_utils.h
```diff
@@ -439,4 +445,51 @@ std::string get_cherrytree_lang_filepath();
...
+    }
+
+    friend path operator/(const path& lhs, const std::string& rhs) {
+        return path(Glib::build_filename(lhs._path, rhs));
+    }
+
+    friend path operator+(const path& lhs, const path& rhs) { return path(lhs._path + rhs._path); }
+    
+    const char* c_str() const { return string().c_str(); };
+    std::string string() const { return get_proper_platform_filepath(_path); }
```
giuspen, line 481: Is .string() used from std::filesystem? Otherwise I would rather prefer .str() being shorter

giuspen:
Eventually Glib::ustring has to go from the file paths completely, you don't have to, but the more you replace with std::string the better.

ForeverRainbow:
I think it would be better to make that a separate pr since otherwise that muddies the history (this is meant to be for custom link handlers xD)

giuspen:
Sure thanks! Please see my second review, in particular the translatable sentences are not to be lost. I'll let @txe do the merge then as soon as he can.

ForeverRainbow:
Can't see a second review? But yeah I will put back the `_()`

future/src/ct/ct_actions_others.cc
ForeverRainbow: `filepath` is a `Glib::ustring` which I didnt change as I mentioned previously

future/src/ct/ct_misc_utils.h
ForeverRainbow: Yes `string()` is from `std::filesystem`

giuspen:
That's fine thanks for your work I'll let @txe do the merge

giuspen:
@ForeverRainbow if you planned to do something on top of this PR and this merge is stopping you please let me know so I'll merge anyway and @txe will double check later

future/src/ct/ct_misc_utils.h
```diff
@@ -439,4 +445,51 @@ std::string get_cherrytree_lang_filepath();
...
+
+    friend path operator/(const path& lhs, const path& rhs) {
+        return path(Glib::build_filename(lhs._path, rhs._path));
+    }
+
+    friend path operator/(const path& lhs, const std::string& rhs) {
+        return path(Glib::build_filename(lhs._path, rhs));
+    }
+
+    friend path operator+(const path& lhs, const path& rhs) { return path(lhs._path + rhs._path); }
```
txe, line 478: what the purpose of operator + ? Why not use Glib::build_filename here?

ForeverRainbow: Suppose I have `foo/bar` and I want to add `.txt` to the end, afaik `Glib::build_filename` will add a seperator instead of `foo/bar.txt` I get `foo/bar/.txt`?

txe: I see, it's better to use std::filesystem approach `+= string`, because current `path + path` is confusing

txe:
Thanks

PR #928, opened 2020-06-17
concurrently convert images into string before node saving

### LINKED ISSUE #854: Cherry Tree Freezing up on Save of Large Notes
I'm using Cherrytree for class notes taken from videos.  My notes include lots of screen-shots.  When a note gets big, the time to save gets very long, during which Cherrytree freezes up.  Then, I find myself setting preferences to save it less and less frequently.

1) Is it possible that this can somehow be done in the background so I am not blocked during the save?

2) Is it possible to have the setting in preferences based not just on time, but on the size of changes or the number of changes in the note since the last save?  So just off the top of my head, something like this:

```
if (note size has changed by X bytes)
or (>= Y characters have been typed)
or (>= T time has passed)
then (save)
```

txe:
Resolves #854

**Issue**: saving pixbuffers into buffers is slow. Many images (e.g. when nodes are imported, or edited node has a lot of images) noticeably slow down saving.
**Solution**: gather all pixbuffers and save them into buffers concurrently. Pass saved buffers in `CtStorageCache` (which can be skipped, then code will work as it worked before)

Notes:
- `tbb::parallel_for` is ideal here, but that adds `libtbb-dev` dependency. Concurrent STL also uses `tbb`, and for some reason it doesn't parallel calculation, I don't know why. Glib has some task classes, but they are really complicated and I concur they have different purpose.
I wrote `parallel_for` by using `std::thread` and `std::packaged_task`. The function is covered by unit tests.
- on my machine, 50 images: old code - 16 seconds; new code - 4.3 seconds.
- improvement works for sqlite and xml

Someone comments?

### DIFF
--- a/future/src/ct/ct_clipboard.cc
+++ b/future/src/ct/ct_clipboard.cc
@@ -194,7 +194,7 @@ void CtClipboard::_paste_clipboard(Gtk::TextView* pTextView, CtCodebox* /*pCodeb
 void CtClipboard::table_row_to_clipboard(CtTable* pTable)
 {
     CtClipboardData* clip_data = new CtClipboardData();
-    pTable->to_xml(clip_data->xml_doc.create_root_node("root"), 0);
+    pTable->to_xml(clip_data->xml_doc.create_root_node("root"), 0, nullptr);
     clip_data->html_text = CtExport2Html(_pCtMainWin).table_export_to_html(pTable);
 
     _set_clipboard_data({TARGET_CTD_TABLE, TARGETS_HTML[0]}, clip_data);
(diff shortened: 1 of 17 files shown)


giuspen:
Interesting work, it's not clear to me what the unit test is demonstrating, some comments there would help

txe:
`parallel_for` splits the given range on slices, one slice per thread. The formula to calculate slices is not simple, so the unit test checks that every element in the given range is processed, processed only once, no one is skipped, and elements that are out of the range should not be touched. To be sure, it is checked for different combinations of ranges, including empty range.

giuspen:
Ok thanks @txe

PR #938, opened 2020-06-18
Refractor CtFileSystem

ForeverRainbow:
This is a refractor of CtFileSystem and changing everything to use `CtFileSystem::path`, afaik there are no more `Glib::ustring`'s being used for paths (hazzah!). 

I have added unit tests in `tests_filesystem.cpp` and currently everything is passing on linux at least

(Also yes I misclicked when writing this, oops)

### DIFF
--- a/future/src/ct/ct_filesystem.cc
+++ b/future/src/ct/ct_filesystem.cc
@@ -31,7 +31,7 @@
 #include "ct_logging.h"
 #include "ct_config.h"
 
-namespace CtFileSystem {
+namespace fs {
 void path::append(const path& other)
 { 
     _path = Glib::build_filename(_path, other._path); 
@@ -39,10 +39,15 @@ void path::append(const path& other)
 
 path path::extension() const
 {
-    auto last_pos = _path.find('.');
-    if (last_pos == std::string::npos) return path("");
-    else                               return path(_path.begin() + last_pos, _path.end());
+    std::string name = filename().string();
+    auto last_pos = name.find_last_of('.');
+    if (last_pos == std::string::npos || last_pos == name.size() - 1 || last_pos == 0) {
+        return path("");
+    } else {
+        return path(name.begin() + last_pos, name.end());
+    }
 }
+
 
 path& path::operator=(path::string_type other)
 {
@@ -52,7 +57,7 @@ path& path::operator=(path::string_type other)
 
 
 
-bool remove(const CtFileSystem::path& path)
+bool remove(const fs::path& path)
 {
     if (fs::is_directory(path)) {
         // rmdir
@@ -130,10 +135,10 @@ std::uintmax_t file_size(const path& path)
     return st.st_size;
 }
 
-std::list<CtFileSystem::path> get_dir_entries(const path& dir)
+std::list<fs::path> get_dir_entries(const path& dir)
 {
     Glib::Dir gdir(dir.string());
-    std::list<CtFileSystem::path> entries(gdir.begin(), gdir.end());
+    std::list<fs::path> entries(gdir.begin(), gdir.end());
     for (auto& entry: entries)
         entry = dir / entry;
     return entries;
--- a/future/src/ct/ct_filesystem.cc
+++ b/future/src/ct/ct_filesystem.cc
@@ -1,0 +1,337 @@
+/*
+ * ct_filesystem.cc
+ *
+ * Copyright 2017-2020 Giuseppe Penone <giuspen@gmail.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+ * MA 02110-1301, USA.
+ */
+
+#include <glibmm/miscutils.h>
+#include <glib/gstdio.h>
+#include <curl/curl.h>
+#include <spdlog/fmt/bundled/printf.h>
+#include <system_error>
+
+#include "ct_filesystem.h"
+#include "ct_misc_utils.h"
+#include "ct_const.h"
+#include "ct_logging.h"
+#include "ct_config.h"
+
+namespace CtFileSystem {
+void path::append(const path& other)
+{ 
+    _path = Glib::build_filename(_path, other._path); 
+}
+
+path path::extension() const
+{
+    auto riter = _path.crbegin();
+    while (riter != _path.crend() && *riter != '.') {
+        ++riter;
+    }
+    if (riter == _path.crbegin()) return path("");
+
+    return path(riter.base(), _path.cend());
+}
+
+path& path::operator=(path::string_type other)
+{
+    _path = _get_platform_path(std::move(other));
+    return *this;
+}
+
+
+
+bool remove(const CtFileSystem::path& path)
+{
+    if (fs::is_directory(path)) {
+        // rmdir
+        int retr = g_rmdir(path.c_str());
+        if (retr != 0) {
+            // error
+            throw fs::filesystem_error("Error occurred during g_rmdir", path, std::error_code(retr, std::generic_category()));
+        }
+        g_unlink(path.c_str());
+    }
+    return true;
+}
+
+bool is_regular_file(const path& file)
+{
+    return Glib::file_test(file.string(), Glib::FILE_TEST_IS_REGULAR);
+}
+
+
+std::string path::_get_platform_path(std::string filepath)
+{
+#ifdef _WIN32
+    filepath = str::replace(filepath, CtConst::CHAR_SLASH, CtConst::CHAR_BSLASH);
+#else
+    filepath = str::replace(filepath, CtConst::CHAR_BSLASH, CtConst::CHAR_SLASH);
+#endif
+    return filepath;
+}
+
+bool copy_file(const path& from, const path& to)
+{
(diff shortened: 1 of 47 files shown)


ForeverRainbow:
Tests failing on windows because the paths are being replaced properly (ironically), fixing that now

giuspen:
Very nice work indeed

txe:
Agree, though there are few issues. I write them down when finish reviewing.

giuspen:
Thanks @txe!

future/src/ct/ct_filesystem.cc
```diff
@@ -0,0 +1,341 @@
...
+        // https://stackoverflow.com/questions/42442189/how-to-open-spawn-a-file-with-glib-gtkmm-in-windows
+#ifdef _WIN32
+        ShellExecute(NULL, "open", folderpath.c_str(), NULL, NULL, SW_SHOWDEFAULT);
+#elif defined(__APPLE__)
+        std::vector<std::string> argv = { "open", folderpath.string() };
+Glib::spawn_async("", argv, Glib::SpawnFlags::SPAWN_SEARCH_PATH);
+#else
+        fs::path path("file://");
+        path += folderpath;
+        g_app_info_launch_default_for_uri(folderpath.c_str(), nullptr, nullptr);
```
txe, line 197: folderpath -> path

future/src/ct/ct_filesystem.cc
```diff
@@ -0,0 +1,341 @@
...
+            count += remove_all(file);
+        }
+        remove(file);
+    }
+    remove(dir);
+    ++count;
+    return count;
+}
+
+std::string get_cherrytree_datadir()
```
txe, line 236: why not fs::path?

txe: I see `fs::path get_cherrytree_localedir()` with fs::path, but others are std::string

future/src/ct/ct_misc_utils.cc
```diff
@@ -760,6 +728,17 @@ char* CtRgbUtil::set_rgb24str_from_str_any(const char* rgbStrAny, char* rgb24Str
     return rgb24StrOut;
 }
 
+std::vector<std::string> CtMiscUtil::split(const std::string& str, std::string_view sep) {
```
txe, line 731: there is `str::split`, why one more?

future/src/ct/ct_misc_utils.h
```diff
@@ -47,14 +47,22 @@ namespace CtCSV {
 }
 
 namespace CtMiscUtil {
+template<class ITER_T>
```
txe, line 50: there is `str::join`

future/src/ct/ct_filesystem.cc
ForeverRainbow: Must have missed those, I'll correct

future/src/ct/ct_misc_utils.cc
ForeverRainbow: Because I didn't know about `str::split` is why xD

txe:
Another big thing is using `_get_platform_path`, there are a few reasons why it's bad:
- it looks like a sign of bad code when it is applied every time to every path. There were only a few places where `get_proper_platform_filepath` was used, though I don't see which problem it was fixing.
- it actually breaks paths, from this PR
```
fs::path path("file://");
path += folderpath;
```
on win `//` will be replaced by `\\`. Also paths can contain network paths, they will be also broken.

I suggest to remove `_get_platform_path` and revert `get_proper_platform_filepath` (still don't know what they are for)


the code for `path path::extension() const` doesn't look good, can you rewrite it by using the same approach as 'stem'?  Also about `stem`, it returns `path` though `string` have more sense.

ForeverRainbow:
`stem()` returns `path` because it could very easily be a `path`, consider `/foo/bar.txt`, `stem() -> /foo/bar` which is a path. Also the specification states that `stem()` returns a `path` object. I have avoided network paths using `path` precisely for that reason (I think I have used it in a few places which is a problem).

The problem with a separate function is it breaks the guarantee that `string()` returns a correctly formatted path, which is specified by the standard. Our version of `path` doesn't have to be completely compatible with the standard of course, it just might cause confusion if later we switch to a compatible one. Originally `get_proper_platform_filepath` was called on every call to `string()` but I moved it. Instead of making it separate I could move it back or I could specialise `CtSplittable` to build a path using native separators and use that?

ForeverRainbow:
I think calling `_get_platform_path` would probably be better since otherwise theres just the issue again of formatting when not needed. The standard does define a `format` enum which can be `native`, `generic` or `auto` (implementation define), `generic` just uses `/` seperators whereas `native` is what we have now.

ForeverRainbow:
Yeah I will revert to `get_proper_platform_filepath`, the only places I think its actually needed are with the `system` calls since powershell can handle posix paths anyway (but `std::system` used `cmd.exe` even on windows 10 because reasons) but I will do it tomorrow

giuspen:
What is your timezone @ForeverRainbow?

txe:
well, `/foo/bar.txt, stem() -> /foo/bar` is not correct,  `stem()` should return `bar`, you can check `std::fs::path` specification

ForeverRainbow:
@txe Yeah your right it should (will check that this one does), so it would make a bit more sense to return `string()` but since the standard specifies `path` it would make life easier for swapping it out later. 

@giuspen GMT

ForeverRainbow:
OK I have moved the `_get_platform_path` call to `native()`, really `string()` should return `native()` (according to the standard) but since the change is compatible with any standard API it doesn't matter afaik (we just have to remember that `string()` does not return a windows correct path on windows). I think I have changed it everywhere it matters

giuspen:
sorry @txe what is GMT?

giuspen:
OK @ForeverRainbow I got the GMT, I thought it was written from @txe and didn't understand the meaning

txe:
Updated `extension()` doesn't work right

txe:
@giuspen , if namespace `fs` is mostly used everythere, maybe renaming CtFileSystem to `fs` is a logic step forward?

giuspen:
@txe absolutely yes

future/src/ct/ct_storage_control.cc
```diff
@@ -158,17 +159,17 @@ bool CtStorageControl::save(bool need_vacuum, Glib::ustring &error)
         if (need_backup)
         {
             // move is faster but the file is used by sqlite without encrypt
-            if (_file_path == _extracted_file_path && CtMiscUtil::get_doc_type(_file_path) == CtDocType::SQLite)
+            if (_file_path == _extracted_file_path && CtFileSystem::get_doc_type(_file_path) == CtDocType::SQLite)
             {
                 _storage->close_connect();    // temporary, because of sqlite keepig the file
                 if (!CtFileSystem::copy_file(_file_path, main_backup))
-                    throw std::runtime_error(str::format(_("You Have No Write Access to %s"), Glib::path_get_dirname(_file_path)));
+                    throw std::runtime_error(fmt::format(_("You Have No Write Access to {}"), _file_path.parent_path()));
```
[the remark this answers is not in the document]
ForeverRainbow: Does `fmt` not respect the translated strings? Looking at `str::format` it looks like it just replaces `%s` with `{}` and then runs it through `fmt`? Or is the `%s` needed for translation?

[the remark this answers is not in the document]
txe: %s needed for translation. Basically, any text in _() must not be changed

giuspen:
@ForeverRainbow the _("%s") is part of the patter to be translated so you will temporarily lose the match thus that translation but you don't have to worry, the translators will fix this before the first release

giuspen:
Sorry for the confusion I'm happy to keep the %s for now as @txe advises

future/tests/tests_filesystem.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+TEST(FileSystemGroup, is_directory)
+{
+    CHECK_EQUAL(true, CtFileSystem::is_directory(unitTestsDataDir));
+    CHECK_EQUAL(true, CtFileSystem::is_directory(CtFileSystem::path(unitTestsDataDir).parent_path()));
+}
+
+TEST(FileSystemGroup, get_doc_type)
+{
+    CHECK(CtDocType::SQLite == CtFileSystem::get_doc_type(unitTestsDataDir + "/test.ctb"));
```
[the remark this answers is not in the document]
ForeverRainbow: Which is the 4th? I can only find `None`, `SQLite`, or `XML`?

ForeverRainbow:
`fmt` has got a `sprintf` method btw, not sure if that helps with the translation but `str::format` could use `fmt::sprintf`?

future/tests/tests_filesystem.cpp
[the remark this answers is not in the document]
txe: .ctb  .ctx  .ctd .ctz

txe:
> fmt has got a sprintf method btw, not sure if that helps with the translation but str::format could use fmt::sprintf?

If it understands %s, then yes

ForeverRainbow:
@txe https://fmt.dev/latest/api.html?highlight=sprintf#_CPPv2N3fmt7sprintfERK1SDpRK4Args only problem is I don't think it works with user-defined types (i.e `Glib::ustring`)

txe:
I think we can just fix all translated string by replacing %s by {} in code and translation files. So for now we can left str::format as it is

future/tests/tests_filesystem.cpp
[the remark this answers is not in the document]
ForeverRainbow: Sorry your comments are lagging for me, only just got that

txe:
The tests for extension looks really great :)
So far code looks OK, but it is hard to say by just reviewing diffs.
I'm going to take your branch and look through by IDE. Maybe run code on win. It takes this day.

txe:
Hi, there are some issues,
Maybe the process will go faster if I add myself changes in your branch?

ForeverRainbow:
@txe Sure that would be great!

txe:
remove exceptions because
- they don't help much in this case and code actually doesn't catch them
- it's confusing, a couple functions throw exceptions but others not

remove fs::append, operator path::/= 
- they don't behave like std::filesystem analogs
- std::filesystem analogs are really confusing, it's better to avoid using them at all

PR #940, opened 2020-06-19
fixes gtk not sending paired released Ctrl for pressed Ctrl (broken zoom)

### LINKED ISSUE #889: Ver 0.39.3 has issues with zooming (ctrl+mouse wheel). wont release zoom after releasing ctrl key
also ctrl+touchpad up/down results in cherrytree locked into zoom adjust mode and wont release. Only option is to restart the application every time you need to change the zoom ratio. 

OS: Debian and LinuxMint DE
DE: Cinnamon 4.4.8

txe:
Resolves #889
Fixes situation when you press Control button, we get Control signal; when you release this Control button, we get ISO_Prev_Group instead of Control (LMDE 4, Debian 10)

It only affects pygtk, C++ version uses different approach.


giuspen:
Great thanks

Issue #926, opened 2020-06-16
Enable system tray docking on GNOME >= 3.26 doesn't work properly


ForeverRainbow:
GNOME removed status icons in 3.26, which means that if you are running a version of GNOME >= 3.26 the "Enable system tray docking" option will cause cherrytree to effectively go zombie and "Start Minimized in the system tray" will visually do nothing, requiring the config file to be deleted before cherrytree will start properly.

Since enabling options which prevent you from opening the app is generally not a good idea, I think that there should at least be a warning if this option is enabled on a system running a version of gnome without status icons.

giuspen:
This issue exists also with the current Pygtk2 version, not only with GNOME but other linux desktops, maybe there is a way to know that the system tray is not supported in the current desktop and grey out the checkbox in that case leaving appindicator as only option.

ForeverRainbow:
Looks like [gtk_status_icon_is_embedded](https://developer.gnome.org/gtk3/stable/GtkStatusIcon.html#gtk-status-icon-is-embedded) can be used to check for the existence of a system tray, the deprecated warning is just for using it for notifications I think

ForeverRainbow:
Also you can use extensions to bring back the gnome tray so completely disabling this might also not be the best but :woman_shrugging: 

I am assuming there is no way to replicate the status icon functionality with `GNotification` (since it looks to be only dismiss-able notifications)

txe:
Was resolved by #941

ForeverRainbow:
@txe This wasn't resolved in the python version though, #941 only patches the C++ version

txe:
it's OK. I think the C++ ver. is more or less ready to be a replacement for pygtk.

PR #941, opened 2020-06-19
Fix being able to minimise to systray on systems without a systray

ForeverRainbow:
This is a fix for #926 on the C++ version. It disables the option to enable docking to the system tray and forces normal startup on desktop environments without system trays (e.g gnome). 

I have tested it on a windows 10 machine and the system tray works fine there (the option isnt disabled and it minimises)

### DIFF
--- a/future/src/ct/ct_main_win.cc
+++ b/future/src/ct/ct_main_win.cc
@@ -510,7 +510,7 @@ void CtMainWin::config_apply()
     _ctStatusBar.stopButton.hide();
 
     get_status_icon()->set_visible(_pCtConfig->systrayOn);
-    menu_set_visible_exit_app(_pCtConfig->systrayOn);
+    menu_set_visible_exit_app(_pGtkStatusIcon->is_embedded() ? _pCtConfig->systrayOn : false);
 
     _ctTextview.set_show_line_numbers(get_ct_config()->showLineNumbers);
     _ctTextview.set_insert_spaces_instead_of_tabs(get_ct_config()->spacesInsteadTabs);
--- a/future/src/ct/ct_main_win.cc
+++ b/future/src/ct/ct_main_win.cc
@@ -124,10 +124,17 @@ CtMainWin::CtMainWin(bool             start_hidden,
     menu_set_items_special_chars();
     _uCtMenu->find_action("ct_vacuum")->signal_set_visible.emit(false);
 
-    if (start_hidden || (_pCtConfig->systrayOn && _pCtConfig->startOnSystray))
-        set_visible(false);
-    else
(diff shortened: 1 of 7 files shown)


future/src/ct/ct_main_win.cc
```diff
@@ -124,10 +124,17 @@ CtMainWin::CtMainWin(bool             start_hidden,
     menu_set_items_special_chars();
     _uCtMenu->find_action("ct_vacuum")->signal_set_visible.emit(false);
 
-    if (start_hidden || (_pCtConfig->systrayOn && _pCtConfig->startOnSystray))
-        set_visible(false);
-    else
+    if (start_hidden || (_pCtConfig->systrayOn && _pCtConfig->startOnSystray)) {
```
txe, line 127: start_hidden is used to hide windows during import from command line. It's not related to tray, so it has to work

txe:
'menu_set_visible_exit_app' should be called with false when no systray

ForeverRainbow:
@txe If there is no systray then it will be anyway? Since the user cannot enable the option?

txe:
all menu items are created visible by default. By calling this function, we control visibility of App Exit (menu item); we should hide it if not systray

future/src/ct/ct_pref_dlg.cc
```diff
@@ -1343,6 +1343,11 @@ Gtk::Widget* CtPrefDlg::build_tab_misc()
     Gtk::VBox* vbox_system_tray = Gtk::manage(new Gtk::VBox());
     Gtk::CheckButton* checkbutton_systray = Gtk::manage(new Gtk::CheckButton(_("Enable System Tray Docking")));
     Gtk::CheckButton* checkbutton_start_on_systray = Gtk::manage(new Gtk::CheckButton(_("Start Minimized in the System Tray")));
+    bool has_systray = _pCtMainWin->get_status_icon()->is_embedded();
+    checkbutton_systray->set_sensitive(has_systray);
+    if (!has_systray) {
+        checkbutton_systray->set_tooltip_text("Your system does not appear to support system trays");
```
giuspen, line 1349: this should be translated, so set_tooltip_text(_("Your system does not appear to support system trays"));

txe:
Now I see why you asked, you skipped another  `menu_set_visible_exit_app`

ForeverRainbow:
Where? Couldn't find it in `ct_main_win`

txe:
in `void CtMainWin::config_apply()`, line before the last one

ForeverRainbow:
Sorry, had to do a thing. OK fixed it.

txe:
That's OK, thanks

PR #944, opened 2020-06-20
copy node link to clipboard; copy anchor link to clipboard

### LINKED ISSUE #7: RFE: simpler way to create links to other nodes
```
This is a request for enhancement.

At the moment, the process to create a link to another node is as follow:
1. Put your cursor in the location where you want the link to be inserted
2. Type Ctrl+L (or use the menu)
3. Type in the name of the link, usually it will be the name of the node you 
want to link to. 
4. On the popup, select "To Node"
5. Navigate to the node you want to link to and press OK.

Issues:
1. you have to type and remember the name of the target node before the link is 
inserted.
2. the process is long


There might be several ways to address this. Here's one way which I think is 
not too bad, that's how OneNote solves the issue:
On each linkable objects (which could be only nodes for Cherrytree at first), 
provide an entry in the context menu labeled "Copy link to node".
Then, you can simply use the "Paste" action (ctrl+V, item from the context 
menu, ...) to paste the link to the other node.

The mechanism might not be immediately discoverable if you don't know about it, 
but once you know it is very efficient to create links to other objects. It is 
also generic if you want to create links to paragraphs, or links to other 
cherry tree documents and so on.


```

Original issue reported on code.google.com by `sylvain....@gmail.com` on 20 Oct 2014 at 6:13

txe:
Resolves #7

@giuspen , feel free to change menu item names and add icons :)

### DIFF
--- a/future/src/ct/ct_actions.h
+++ b/future/src/ct/ct_actions.h
@@ -181,6 +181,7 @@ public:
     void tree_sort_ascending();
     void tree_sort_descending();
     void tree_info();
+    void node_link_to_clipboard();
     void node_siblings_sort_ascending();
     void node_siblings_sort_descending();
     void node_go_back();    // was as go_back
@@ -347,6 +348,7 @@ public:
     void anchor_copy();
     void anchor_delete();
     void anchor_edit();
+    void anchor_link_to_clipboard();
     void embfile_cut();
     void embfile_copy();
     void embfile_delete();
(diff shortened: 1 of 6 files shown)


giuspen:
The menu item names are good, the only thing in one you've used capital first letters and in the other not. I've also not been consistent with that actually, we should decide in futore one mode and stick to it and gradually fix the others

txe:
It was a mistake, I tried  to make every work capital excluding `to` and articles

txe:
Should wait a review(

giuspen:
It's me that shall start to ask for your review when working on bigger tasks... you have the whole picture while I don't in this busy moment 😉

txe:
Let's hope there is no need :), I already feel like I'm working a fulltime job here this weeks, so many emails/answers/reviews

giuspen:
😁 sure no worries

PR #954, opened 2020-06-22
Fix issues with md formatting

ForeverRainbow:
This fixes an issue raised [here](https://github.com/giuspen/cherrytree/issues/444#issuecomment-647577303). Basically the parser was overzealous with erroring out.

To note, the formatter is still not perfect (case in point: tables are problematic since it is done line by line, but since that's more advanced I don't think it should be expected to work). This also removes formatting on space, replacing it with only on enter since one space is pretty restrictive. Perhaps there should be a config option for when to do formatting?

Also this adds `_` as italics (forgot that part of the syntax, oops)

### DIFF
--- a/future/src/ct/ct_widgets.cc
+++ b/future/src/ct/ct_widgets.cc
@@ -458,10 +458,11 @@ void CtTextView::for_event_after_key_press(GdkEvent* event, const Glib::ustring&
                 start_iter.backward_line();
                 try {
                     _markdown_check_and_replace(text_buffer, start_iter, iter_insert);
-                    text_buffer->insert_at_cursor(CtConst::CHAR_NEWLINE);
                     iter_start  = text_buffer->get_insert()->get_iter();
                     iter_insert = iter_start;
-                } catch(CtParseError&) {}
+                } catch(const CtParseError& e) {
+                    spdlog::error("Exception caught while parsing markdown formatting: {}", e.what());
+                }
             }
             
             int cursor_key_press = iter_insert.get_offset();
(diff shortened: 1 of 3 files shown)


future/src/ct/ct_widgets.cc
```diff
@@ -616,19 +617,6 @@ void CtTextView::for_event_after_key_press(GdkEvent* event, const Glib::ustring&
                     if (iter_start.get_line_offset() == 0 and iter_start.get_char() == g_utf8_get_char(CtConst::CHAR_COLON))
                         // ":: " becoming "▪ " at line start
                         _special_char_replace(CtConst::CHARS_LISTBUL_DEFAULT[2], iter_start, iter_insert);
-                } else if (_pCtMainWin->get_ct_config()->enableMdFormatting && syntaxHighlighting == CtConst::RICH_TEXT_ID) {
```
txe, line 619: I wonder, you decided to remove real time formatting?

ForeverRainbow: Remove it after space, it still activates on enter (As I mentioned in the description)

ForeverRainbow: It should possibly be a config option, but I found when using it that space was very annoying when trying to edit stuff

txe: Ah, I see, for some reason, I thought removing formatting meant converting)

txe:
Well, maybe it's a best place to write about it, but ok

I didn't try md formatter yet, but I can take your word that applying md formatter after space can be irritating for user. On the other hand, applying md formatter after return can be confusing because it gives a big delay between writing tags and seeing result after pressing return.

I have an idea how to combine it, though it maybe not simple technically. You can see such behavior in Discord.
The idea to apply md formatter after every pressed key for the current line. If there are tags to apply, we apply styles, but NOT removing tags. So user can even add more tags or removed previous tags and see results instantly. When user starts a new line or press enter, we remove tags and leave styles as they are.

ForeverRainbow:
@txe I was thinking of something like that, the issue is that it would require some storage of md tag storage and such within the widget and would be not massively complex but not trivial. I am assuming you were thinking something like formatted text tag applied with the markdown parts still showing and then remove them on enter. I guess open and close tags could be matched and then marks used to find the parts to remove but yes, not trivial

The problem is as well that it matches some tags its not supposed too (e.g `| row | ` will cause `| row |` to vanish and no table to happen, since its single line).

kettenfett:
#957

PR #962, opened 2020-06-25
xml parser can't parse some symbols so trim them off

### LINKED ISSUE #955: CherryTree stopped saving.
So after restarting CherryTree for the first time in a few days, I've just found that it had not actually saved a thing for 4 days despite manual and auto saves, and the presence of files with dates indicating saves just before restarting CherryTree.

This is on Windows 10 x64, CherryTree 0.39.3. I'm assuming I've just lost a good 24 hours worth of work - but if there's some log or some way I can debug this so it doesn't happen to other people, I'm happy to help out, but to say I'm annoyed right now would be a pretty big understatement.

The files are encrypted sql files (.ctx).

txe:
Resolves #955

### DIFF
--- a/future/src/ct/ct_clipboard.cc
+++ b/future/src/ct/ct_clipboard.cc
@@ -444,7 +444,7 @@ void  CtClipboard::_on_clip_data_get(Gtk::SelectionData& selection_data, CtClipb
 // From Clipboard to Plain Text
 void CtClipboard::_on_received_to_plain_text(const Gtk::SelectionData& selection_data, Gtk::TextView* pTextView, bool force_plain_text)
 {
-    Glib::ustring plain_text = selection_data.get_text();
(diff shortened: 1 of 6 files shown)


giuspen:
Thanks @txe . Interesting that I have never clearly understood why this filter was required in the pygtk2 version, I copied don't remember where, and was waiting for the first issues to see if we needed it in the Gtkmm3 version as well and the answer is here. If you find out additional info as to why these characters are to be filtered out let me know (but no need to waste time on it since nobody ever complained for that filter).

txe:
> When you put utf-8 encoded strings in a XML document you should remember that not all utf-8 valid chars are accepted in a  XML document http://www.w3.org/TR/REC-xml/#charsets
So you should strip away the unwanted chars, else you’ll have an XML fatal parsing error such as above

giuspen:
There you go, thanks!

Issue #858, opened 2020-05-28
Implement markdown parsing and formatting


ForeverRainbow:
Hi
This is an attempt to provide somewhere for further discussion on the markdown parsing mentioned in #828. It is something I would like as well as I am used to using markdown for formatting my notes and general writing.

Right now I think the implementation of markdown falls into two categories: Importing/exporting, and formatting. The import/export category is pretty self-explanatory - add the ability to import and export markdown files to cherrytree. The formatting catagory is more complex, if cherrytree does get the ability to interpret markdown (for example to write \_italics\_), then how much of the "standard" should it support and what (if any) conflicts would it cause with the existing shorthand language?

txe:
Yeah, actually, I wanted to asked it too)
1. which standard to support? should we support several standards?
2. how much. I don't use markdown, so you can come up with anything reasonable. And we can outline which parts don't make sense in cherrytree e.g. lists.
3. About conflicts. I don't see so far. And definitely there will be an option to turn off markdown. 
 
So based on github [markdown](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#headers) there are

Headers - yes (one guy asked about it long time ago)
Emphasis - yes
Lists - no
Links - yes
Images - yes 
Code and Syntax Highlighting - yes
Tables - no
Blockquotes - no
Inline HTML - no
Horizontal Rule - no
Line Breaks - no
YouTube Videos - yes (everyone likes videos with cats 🐱)

txe:
Other issues
- users may want to escape markdown symbols. maybe we cannot help with that.
- paste from clipboard can contain markdown symbols or parts of them (like singular '_')

That's why I want to see editors with real-time markdown support, discord is quite good example, github is not because it has a viewer.

ForeverRainbow:
Escaping markdown is relatively simple - just check if there is a `\` before the symbol. Most editors I know about don't do real-time parsing of the markdown, since the person writing it can generally understand what it means and its very simple (at least thats my guess). Possibly in the configuration there might be an option to have it always trigger when it detect a valid expression (like header text), never trigger, or trigger on some action (e.g hitting enter at end of line)

txe:
> Escaping markdown is relatively simple - just check if there is a \ before the symbol

[...]

PR #963, opened 2020-06-25
Improve realtime markdown formatting

ForeverRainbow:
This is an improvement for realtime markdown formatting. It also adds some bits of syntax which I had forgotten :p

Markdown formatting will now be applied automatically when valid tags are matched. The tags will be removed and the cursor placed at the end of the newly formatted block. So \_italic\_ would become _italic_ as soon as the last \_ was typed. Currently there is no way to undo this, `undo` will remove the entire format block at once. I am not sure how undo actually works so I am assuming there is a way around this but not sure how to implement it. 

Tables have been disabled due to not being able to be closed (thereby locking the parser until a restart). They still work when importing. Also headers with `==` and `---` do not work for obviouse reasons.

I also added `<some url>` for links. I am going to do a writeup of what markdown cherrytree supports, should I just place this in `/future`? (can also serve as a demonstraction of the markdown importer)

Ref #858


ForeverRainbow:
Currently this doesnt check if the node is rich text before applying, oops - fixing it now

txe:
I'd like you to change your code related to CtTextView:
- there is a lot of code in CtTextView for markdown, encapsulate it by putting in e.g. CtMardowBuffer. 
- there is a lot `_pCtMainWin->get_ct_config()->enableMdFormatting && _buffer_is_rich_text()`. Add `enable(bool enable)` instead.
- please don't use functions without namespace, e.g. `get_md_tag_name`. Use lambdas or put them in a class.
- `apply_tag` creates CT tags or your custom tags? if it's CT's, then use `get_text_tag_name_exist_or_create`. If custom ones, then I 'm not sure if CT can handle them properly in different situations.
- functions for signals usually have a pattern in the name like `_on_key_press_event`

ForeverRainbow:
`apply_tag` is a helper to apply or create a custom tag yes. I don't think it should cause issues since the names follow the patten `md-formatting-<num>`. If that is going to cause an issue then I will have to rethink how to do this. Also I went with the `_for` patten with the signal handlers since thats what the other ones in the class seem to do (although they are called manually from CtMainWin), so I will change those

ForeverRainbow:
@txe I'm refractoring the markdown stuff into `CtMarkdownFilter` is just attached to an arbratray `Gtk::TextBuffer`, should I keep it in `ct_widgets` or move it (e.g into `ct_parser`)?

txe:
well, ct_parser sounds good

txe:
I will need to test changes before emerging them

giuspen:
Doesn't build on Mac OS https://travis-ci.com/github/giuspen/cherrytree

giuspen:
@txe I have not looked at this because you said that you were going to test it in a previous comment but if you are busy and/or would prefer me to look at it instead I will try and give feedback to @ForeverRainbow myself as soon as get some time

txe:
Thanks, it would be great!

giuspen:
@ForeverRainbow I'm sorry I kept your PR in standby so long that it started to have conflicts. It's good to go for me though so if you resolve the conflicts I'm going to merge straight away

ForeverRainbow:
Conflicts seem to be purely consmetic, weird

giuspen:
Thanks!

PR #971, opened 2020-06-30
Fix crash macos

txe:
TestCtApp causes a crash on macos. It's not easy to run debugger on macos, so temporary disable it.


giuspen:
Thanks @txe it would be great to understand why it is crashing since these tests are a quite good to ensure the reading and writing is still reliably working for all document types (and I plan to add more methods to verify the existence of the data and the formatting). I wonder if there is a free operative system that somehow mimics the MacOS like CentOS for RedHat. Anyway you made it right I'll try and find out why they're crashing the Mac before re enabling those.

txe:
The issue is not with tests, it is enough to leave only TestCtApp constructor in every test without other code to cause crash. Maybe it related to global vars or something else. I couldn't run gdb on macos, so it will take time to figure out.

giuspen:
It could be because the MacOS server Is running without a graphic server and it is required when the Gtk loop starts. On Linux, the issue was easily solved with xvfb-run wrapping the command (see Travis yaml) maybe there is something equivalent for MacOS

txe:
Though you may be right about travis, but:
- crash is also in my virtual machine with desktop
- the issue is reproduced when several tests use TestCtApp, but if only one or two, tests are OK.

giuspen:
Thanks it is definitely not the graphic server then :) I see a pile of warnings in the MacOS build log, maybe I'll start to try and get rid of some of those and then I'll try again

PR #973, opened 2020-07-01
main.py : don't try libc.prctl on MacOS

kencu:
It's not available on macOS, and so generates a warning every launch.

I'm not a python guy, but this looked right and seems to stop the warning on my systems.


txe:
That's right, thanks!

PR #980, opened 2020-07-02
disable systray on macos for pygtk

txe:


kencu:
Looks so simple when you do it :>

I will test, and patch our version, and thanks. CherryTree is growing on me very nicely. Good luck with it!

Issue #942, opened 2020-06-20
[Feature Request]: Keep links when exporting to pdf


ForeverRainbow:
Hi.
Currently exporting to pdf will result in links being lost, both internal (i.e to other nodes -> exported pages) and external (e.g to a website). 

It would be really useful if links could be exported, since especially when navigating lots of notes, not having my "more info here" links work is quite a big issue. I can always just read these notes in cherrytree of course but its very helpful to be able to distribute them in a more or less universal format like pdf.

Internal and external links are supported in pdf afaik, but I have little to no experience with pdf beyond just using it. I am happy to try and implement this if it is doable though.

ForeverRainbow:
Just to say as well, `Pandoc` does support pdf conversion so that might be an option just for external links, however if internal links are to be kept as well I think something more complex is needed

giuspen:
I was hoping that with Gtk3 or at least Gtk4 the pdf would support links... This feature is quite high in the number of requests so definitely a valid feature to investigate, thanks!

txe:
Cairo supports links
https://www.cairographics.org/manual/cairo-Tags-and-Links.html
https://gitlab.freedesktop.org/cairo/cairo/-/issues/357

ForeverRainbow:
@txe So it would require a Cario dependency? Or does cherrytree already use Cario? Looking at the code for `CtExport2Pdf` it looks like it just translates to pango and then mysteriously a pdf appears (I am guess gtk can create pdfs but cannot find any documentation on it)

txe:
Agree, this stuff is confusing. Both are already used. Cairo is low-level render. Pango is high-level render and uses Cairo.
https://en.wikipedia.org/wiki/Pango

Things are go on in
`CtPrint::_on_begin_print_text`   - to calculate text/images/other positions (layout)
`CtPrint::_on_draw_page_tex`t - is actual drawing

ForeverRainbow:
@txe It looks like the "pipeline" for pdf creation goes something like: `text buffer -> pango -> adjust output positions -> output pango using cairo`. Which means that after the text buffer stage anything not supported in pango markup (i.e links) is lost (with the exception of widgets). 

I am not sure how to get around this without redesigning the pipeline for pdf exporting which would be pretty massive and (maybe?) extreme. Is there anything I am missing here with the export process?

txe:
Yeah, it doesn't look simple.
[...]

PR #981, opened 2020-07-03
Redesign pdf exporting code

ForeverRainbow:
This is the first part to implementing: #942 

I ended up going for the redesign in the end because the custom formatting worked up to a point but was incredibly hacky and trying to get it to align properly and format correctly was a nightmare. 

The new design works as follows:
- `CtPrintData` has a list of `CtPrintable`'s which are everything which can be drawn to the pdf  (i.e text and widgets). 
- Page count is calculated based on the sum heights of all the `CtPrintable`'s in the print data divided by the size of each page (rounded up, +1 if there is any remainder)
- `CtPrintable` exposes `height()`, `width()`, `setup()`, `print()` and `done()`. `setup()` needs to be called before the printable can be used and `print()` displays the printable in the given context and returns the final position of the "cursor" after drawing the printable. `done()` is used to check if the printable is done (e.g the text printable will not be "done" until it has printed all of its line but it may print them over several pages).
- Page breaks are done via `CtPageBreakPrintable` which returns `-1` for the height (< 0 is interpreted as a page break) and moves the position of the cursor out of bounds on `print()`.
- Widget printables inherit from the template `CtWidgetPrintable`, which inherits from `CtPrintable`, based on their widget (e.g `CtWidgetTablePrintable` inherits from `CtWidgetPrintable<CtPrintTableProxy>`), the dynamic casts have been moved to `printable_from_widget()`. It may be possible to avoid the proxy classes all together here but I did not think it was really necessary. 

`_on_begin_print_text` now simply calls `setup()` on all the printables and calculates the number of pages (as described above). 

`_on_draw_page_text` now calls `print()` on the current printable and updates the next position based on the return value, if the printable is done it increments the current printable, if the new position after printing is larger than the height of the page it returns.

The pango serialization methods now all return either `std::vector<std::shared_ptr<CtPrintable>>` (`CtPrintableVector`) or `std::shared_ptr<CtPrintable>`.

Based on this structure it should be simple to implement links and anchors since `print()` gets the current position and a cairo context to use and each `CtPrintable` is completely independent.

### DIFF
--- a/future/src/ct/ct_export2pdf.cc
+++ b/future/src/ct/ct_export2pdf.cc
@@ -21,34 +21,36 @@
 
 #include "ct_export2pdf.h"
 #include "ct_dialogs.h"
+#include "ct_logging.h"
 
 #include <utility>
 
 void CtExport2Pdf::node_export_print(const fs::path& pdf_filepath, CtTreeIter tree_iter, const CtExportOptions& options, int sel_start, int sel_end)
 {
-    std::vector<Glib::ustring> pango_slots;
+    CtPrintableVector printable_slots;
     std::list<CtAnchoredWidget*> widgets;
     Glib::ustring text_font;
     if (tree_iter.get_node_is_rich_text())
     {
-        CtExport2Pango().pango_get_from_treestore_node(tree_iter, sel_start, sel_end, pango_slots, true /*exclude anchors*/, widgets);
+        CtExport2Pango::pango_get_from_treestore_node(tree_iter, sel_start, sel_end, printable_slots, true /*exclude anchors*/, widgets);
         text_font = _pCtMainWin->get_ct_config()->rtFont;
     }
     else
     {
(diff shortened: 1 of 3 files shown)


txe:
Proxy objects like CtPrintTableProxy was used to keep additional information without unnecessary changing widgets.
With CtWidgetPrintable they became unnecessary, and it doesn't have sense to have additional layer.

txe:
`printable_from_widget` is used only in CtExport2Pango, why not to make it private and put it there?

txe:
Though you are right, if you begin remove proxy code, the PR become less clear, so it's unecessary

txe:
How heavy did you test changes? I think of just merging it and fixing some insignificant parts later.

ForeverRainbow:
I tested them with a large tree (214 page for node on each page) and a single node with lots of newlines (to split text over pages). Also tested node with and without node in new page, exporting only selection, node and subnodes, and only this node. 

As far as I can tell the output is identical to the old output (except exporting only selection has an extra page I think).

txe:
Ok, thanks!

PR #982, opened 2020-07-04
Implement links in exported PDFs

ForeverRainbow:
This is the second part to #942

Image anchors are now processed, anchors are inserted as the unicode anchor emoji (⚓) instead of the image because it was simpler, since it takes a `CtImageAnchor` in the relevant ctor however, this is easy to change. 

This adds two new classes, `CtDestPrintable` and `CtLinkPrintable` where `CtDestPrintable` is what processes `CtImageAnchor`'s, and `CtLinkPrintable` just applies the cairo link tag and makes its text blue.

Internal destinations are represented in the form `n.<node id>.[section (e.g h1-1)]`, the section part is optional. External destinations are processed by `CtStrUtil::external_uri_from_internal()` which converts an internal uri (e.g `webs https://example.com`) to an external one (and decodes file paths).

### DIFF
--- a/future/src/ct/ct_export2pdf.cc
+++ b/future/src/ct/ct_export2pdf.cc
@@ -25,6 +25,27 @@
 
 #include <utility>
 
+
+std::string dest_id_from_node_and_anchor(int node_id, const std::string& anchor_name)
+{
+    return fmt::format("n.{}.{}", node_id, anchor_name);
+}
+
+std::string dest_id_from_node_id(int node_id) 
+{
+    return dest_id_from_node_and_anchor(node_id, "");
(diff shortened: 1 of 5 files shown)


txe:
Amazing, thanks!

giuspen:
Indeed this was a very much awaited feature, many thanks!

ForeverRainbow:
I think the windows build of this is still broken

txe:
it's ok, I'll fix it later

txe:
looks like CppUTest is built on Win32 without the Standard C++ Library, so just disabled CHECK_THROWS there

PR #985, opened 2020-07-05
Fix missing CHECK_THROWS on windows

ForeverRainbow:
#984 fixes this by disabling it on windows, this fixes it by copying in the macro from the source (I have linked the line in a comment)


ForeverRainbow:
Ok I am closing this, can't get it to build properly on other systems and it is fixed anyway

txe:
It couldn't build because I had make a mistake. I've fixed it, so you can try PR again.

PR #990, opened 2020-07-06
fixes behaviour systray icon

txe:
- `is_embedded` doesn't work because it should be called in main event loop
- fixed an issue when New Instance could create a hidden window
- changed "Quit" label into "Hide" when it works as hide

Didn't check it on Ubuntu, but I think it should work


txe:
@ForeverRainbow , to keep your informed, 
code, I think, now correctly detect the systray existence.

PR #998, opened 2020-07-08
Fix powershell/perl typo in modules/config.py

Kolossi:
Typo lead to code exec issues for both


txe:
Yes, it was a typo from  #785

txe:
Thanks!

Issue #828, opened 2020-05-23
Compile a list of features missing in the C++ version


ForeverRainbow:
Hi,

I want to help with the porting of cherrytree to C++, however I am not sure what features are important to implement, are a priority, might be easier for someone new, etc. What I think would be really useful is to have a list of features still missing in the C++ version (possibly with the estimated level of difficulty of implementing them). I think this would definitely help any contributors who want to port things over but are unsure as to what to do. 

---
# List of missing features
### Todo
- [ ]  EssentialPIM HTML importing 

### Not planned
- [ ]  KeyNote importing
- [ ]  Knowit importing
- [ ]  Basket importing
- [ ]  TuxCards importing
- [ ]  Treepad importing

### Done
- [x]  Gnote importing
- [x]  KeepNote importing (#1002 but untested)
- [x]  Zim importing (#853)
- [x]  Tomboy importing
- [x]  Mempad importing (#1011)
- [x]  Leo file importing (#1019)
- [x]  RedNotebook importing (Parse exported HTML) (#1026)
- [x]  NoteCase importing (Parse exported HTML) (#1034)

### Todo
- [ ] Progress bar when updating nodes not showing progress
- [ ] Show/Hide the main window
- [ ] Fix naming of node sanity check methods in CtActions
- [ ] Strip bad characters on URI list paste
- [ ] Fix code duplication in `CtCodebox::to_xml` and `CtHtml2Xml::_insert_codebox`
- [ ] Slowness of get_cell in `CtPrintTableProxy`

### Done
- [x] Find a good CSV lib (Wrote own parser #914)
- [x] Custom weblink action (Open weblinks in custom application?) (#925)
- [x] Validate and check corruption of SQLite database when populating the treestore (#943)

### Todo
- [ ] Theme editor (issue #896)

### Done
- [x] Tree summery (Tree Info) - Not too difficult but requires more in depth into the code (4b458f6)
- [x] New version checking - Trivial
- [x] Fix unable to import cherrytree files created by a version before 49e7f9b (when `ts_creation` and `ts_lastsave` were added to the schema) ( #832)
- [x] Table of contents insert (#909)

txe:
Cool, thanks, 
Also, some missing functional can be found be searching 'todo' in code.

giuspen:
1) new version checking - trivial
2) there should be no need of a special utility to add newline in tables at all. the tables framework changed and a table cell is now like a codebox
3) tree summary - not too difficult but requires more in depth into the code
[...]

PR #1002, opened 2020-07-09
Add Keepnote importing

ForeverRainbow:
This adds keepnote importing to the C++ version.

Note I have not tested this, it is simply a rewritten version of the python one and may or may not actually work. For future reference see https://github.com/giuspen/cherrytree/issues/828#issuecomment-656202172


giuspen:
Thanks!

PR #1011, opened 2020-07-14
Implement mempad importing

ForeverRainbow:
This implements mempad importing in the C++ version. 

Due to imported nodes needing a single parent, the importer creates a dummy node called "Mempad Root", of which all imported nodes are children. 

I have moved the inheritance tree for `CtParser` and `CtTextParser` around a bit to facilitate a parser which doesnt use tokens, I do want to refractor `ct_parser` soon if thats ok. The parsing code is quite ugly I'm afraid because that was the only way I could think of dealing with the weird format of Mempad which uses raw bytes along with encoded text :roll_eyes:

I have put my test data in `tests/data/mempad_test.lst`

Ref: #828


giuspen:
Thanks!

PR #1015, opened 2020-07-14
Refactor ct_parser and ct_imports

ForeverRainbow:
This is a refactor of ct_parser and a bit ct_imports. Changes:

- The inheritance tree of CtParser has been replaced by `CtParserInterface` and all the code for building the XML has been moved to `CtDocumentBuilder` which is used by some of the parsers.
- `CtTextParser` is now separate and again is used by some of the parsers
- `CtMDParser` has been moved from `ct_imports` into `ct_parser`, `CtZimParser` has been added which separates the Zim importer from the parser.
- `ct_imported_node` now has an `std::shared_ptr` to its document instead of copying the tree (this was done so that the broken links container could be copied without leaving dangling pointers).
- Needed maps in `CtTextParser` are set up at construction.
- Fixed define check in require: `NODEBUG` -> `NDEBUG`.

Also as a result of separating `ct_imports` and `ct_parser`, changing one no longer triggers recompiling 30+ TUs (I have also added some forward declarations to allow this).


giuspen:
I would prefer if @txe can review this since he ported some imports while I ported none. In the meantime if you can merge into your branch the latest master and push we should see if it builds properly on macos as well...

future/src/ct/ct_clipboard.cc
```diff
@@ -34,6 +34,7 @@
 #include <gio/gio.h> // to get mime type
 #include <glibmm/regex.h>
 #include "ct_logging.h"
+#include "ct_parser.h"
```
txe, line 37: why is it here?

future/src/ct/ct_imports.cc
```diff
@@ -1175,7 +1026,7 @@ void CtZimImport::_ensure_notebook_file_in_dir(const fs::path& dir)
     }
 }
 
-
+CtMDImport::~CtMDImport() = default;
```
txe, line 1029: why destructor in .cc file?

txe:
Do you move 'CtHtml2Xml ' class to ct_parser.h, but leave code in ct_import.cc?

future/src/ct/ct_clipboard.cc
txe: Got it

ForeverRainbow: It needs `CtHtml2Xml`

future/src/ct/ct_imports.cc
ForeverRainbow: `CtMDParser` is forward declared and then used via an `std::unique_ptr`, the default generated dtor doesn't work because it is inline and `std::unique_ptr`'s dtor needs to have access to the complete type

ForeverRainbow:
@txe Yes it looks like I did, whoops, I was moving it to a separate source file but though better of it and reverted it, I'll fix it now

txe:
failed to build on linux

ForeverRainbow:
Windows build has failed but when building 7z for some reason

giuspen:
I'll kick it again

txe:
Thanks!

PR #1018, opened 2020-07-15
addes abitlity to create several toolbars

### LINKED ISSUE #1010: [Feature Request] Double toolbar
It'd be nice having this feature, since one of my monitors is not large enough to hold all the tool icons I want to have easy access to.

Toolbar

txe:
Resolves #1010
- adds a special element "Split Toolbar" in Preference dlg to split and thus to create more toolbars than one. Maybe we need a new icon for it.
- fixes pygtk which doesn't work well with a new toolbar element
- renames functions from `toolbar` to `toolbars`

### DIFF
--- a/future/src/ct/ct_actions.h
+++ b/future/src/ct/ct_actions.h
@@ -235,7 +235,7 @@ private:
 public:
     // view actions
     void toggle_show_hide_tree();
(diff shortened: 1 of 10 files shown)


giuspen:
Thanks @txe I'll have a look for the icon

PR #1019, opened 2020-07-15
Implement Leo file importing 

ForeverRainbow:
This implements Leo file importing for the C++ port. 

The importer does not do any parsing on the input besides parsing the XML to get the hierarchy and names. Note that it does not preserve the order of nodes.

It parses in three stages, the first stage reads the header which provides the hierarchy and names. The second reads the contents nodes. The third links together nodes from stage 1 and 2 using lookup tables based on `tx` and `t` which are the links between content and header nodes, this stage generates a list of `leo_node`. 

The parser generates a list of `leo_node` objects which are then transformed into a single `ct_imported_node` in the import step. It uses a dummy node as the root ("Leo Root")

Ref: #828


giuspen:
Thanks!

preeset:
@ForeverRainbow, 
I am finding myself unable to import my .leo document into cherrytree. I've compared your leo_test.leo file to my own and nothing jumps out at me about why it's not working.

https://github.com/ForeverRainbow/cherrytree/blob/2fc8c4a7b7a62ffefdc9f2fdd65b758bb0df9c97/future/tests/data/leo_test.leo

When I import your .leo file, I am presented with the 'Who is the Parent?' window and thereafter the data is imported correctly. When I do the same with my .leo file, the 'Who is the Parent?' window does not open and nothing seems to happen.

I've uploaded my .leo file here:

https://share.riseup.net/#UsUT9KeWGSJ3j1GKADq3LA

Thanks for your help!

txe:
@preeset, fixed.

Issue #1012, opened 2020-07-14
[CT-0.99] Big codebox breaks PDF Export


txe:
Early version CT-0.99, PDF export doesn't split along codebox into several pages, though it works right in pygtk.
Current version CT-0.99, PDF export is hanging on a long codebox, I suppose, it's after last refactoring

ForeverRainbow:
@txe Yes I think this is my fault, I removed the code for splitting long codeboxes and tables I suppose I didn't put it back :p

ForeverRainbow:
Actually I know exactly what this is, codeboxes and tables will try and move themselves to the next page if the current one cannot fit them, I think it should just print a pdf without anything after a too large codebox/table but if it hangs thats even more of a problem

giuspen:
If a table or codebox is too big for a page it can try and start from the subsequent page if it fits but otherwise should split, at least this is what the Pygtk2 version does, supporting also split on N pages. Both the table and codebox make sense to be longer than a page and we don't want to drop data. I'm not understanding if this was working before @ForeverRainbow rework or not. I can also have a look at this just if it was working before please pay me know I don't want to repeat the work of @txe in case I would go back and find it

txe:
I don't know if it worked before.
I'll take a look at this issue, anyway I wanted to make some changes there.

giuspen:
Ok thanks @txe !

ForeverRainbow:
@txe I was going to look at this but if you are, splitting it over multiple pages should be easy because `print` will keep getting called on a printable until it sets `_done` to true so you can just keep printing the bits which are missing as long as you like

txe:
It's too much broken, @ForeverRainbow, can you fix it all and check other things?
- document with one empty node, options: All Tree, Include Node Name. Result: generate 3 pages
- document with 2 empty nodes, options: All Tree, Include Node Name. Result: generate 3 pages, only one name
- document with 2 empty nodes, options: All tree, New Node on New Page. Result: only one page

ForeverRainbow:
I think the pages problem is due to the floating point arithmatic used to calcuate the number of pages. When testing with ~400 nodes it started to get noticable with 3 pages missing (if I switch out `fmod` for `remainder` it produces 100 more pages than needed)

giuspen:
@ForeverRainbow if you are in difficult with this issue don't worry you can pass me the ball and I'll take it from here

ForeverRainbow:
[...]

PR #1020, opened 2020-07-16
Fix pdf exporting issues after refractor

ForeverRainbow:
This fixes several issues introduced with the refractor of pdf exporting. 
- Codeboxes and tables will now extend over several pages.
- Page breaks are now properly calculated as part of required page number.
- Wrapping is now accounted for when calculating the number of pages.

I have tested this with ~200, ~2000, and ~4000 page documents. Note that page breaks are appended on the end of each page, which causes an extra page on the end of any document using page breaks. I do not see this as much of an issue but it would be a trivial fix if needed.

Fixes: #1012

### DIFF
--- a/future/src/ct/ct_export2pdf.cc
+++ b/future/src/ct/ct_export2pdf.cc
@@ -105,29 +105,22 @@ void CtExport2Pdf::_nodes_all_export_print_iter(const CtTreeIter& tree_iter, con
     {
         node_printables.emplace_back(std::make_shared<CtTextPrintable>(CtExport2Pango().pango_get_from_code_buffer(tree_iter.get_node_text_buffer(), -1, -1)));
     }
+
+    // Push front node name
     if (options.include_node_name) {
-        node_printables.insert(node_printables.cbegin(), generate_node_name_printable(tree_iter.get_node_name(), tree_iter.get_node_id()));
+        node_printables.emplace(node_printables.cbegin(), generate_node_name_printable(tree_iter.get_node_name(), tree_iter.get_node_id()));
     }
-    if (tree_printables.empty()) {
-        tree_printables = node_printables;
+
+    std::shared_ptr<CtPrintable> break_printable;
+    if (options.new_node_page) {
+        break_printable = std::make_shared<CtPageBreakPrintable>();
+    } else {
+        break_printable = std::make_shared<CtTextPrintable>(str::repeat(CtConst::CHAR_NEWLINE, 3));
     }
-    else
-    {
-        if (options.new_node_page)
-        {
-            node_printables.emplace(node_printables.cbegin(), std::make_shared<CtPageBreakPrintable>());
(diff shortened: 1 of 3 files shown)


txe:
Are you to going to fix this?
>It's too much broken, @ForeverRainbow, can you fix it all and check other things?
> -   document with one empty node, options: All Tree, Include Node Name. Result: generate 3 pages
> -   document with 2 empty nodes, options: All Tree, Include Node Name. Result: generate 3 pages, only one name
> -   document with 2 empty nodes, options: All tree, New Node on New Page. Result: only one page

ForeverRainbow:
@txe I didn't explicitly mention it but those are fixed.

txe:
I'm asking because i checked and it still didn't work

ForeverRainbow:
@txe just checked again, it seems node names are broken for empty nodes, thanks for checking it!

ForeverRainbow:
The issue was `pango_get_from_treestore_node` clears its input vector, so the names were getting removed right after being added

txe:
Can you remove the last empty sheet from pdf?

txe:
Thanks, can you fix a bit more?
- line spacing became smaller and it doesn't look good
- right padding in table cell is 0 and it also doesn't look good.
To compare (text from wiki about ssd)

Screenshot_2020-07-17_13-52-25

Screenshot_2020-07-17_13-50-14

ForeverRainbow:
@txe I have fixed the tables text problem (and simplified its code). I am not getting that issue with the overlapping characters that you seem to be getting but I am having problems where long lines are not wrapping properly which I am looking into. Could you upload that test document so I can see what it looks like on my machine?

ForeverRainbow:
Well I've found the cause of the text wrapping issues. The problem is that a `Pango::Layout` is meant to be a full paragraph and lines will not wrap automatically on their own. When links are embedded into the text it breaks up the layout and causes the clipping (since the layout line is drawn from somewhere further acrosss than it expects). I honestly I am not sure how to fix this. Looking at the documentation for pangomm it looks like the only option may be to do it manually using `Pango::Context` and cairomm, but that would be very complex by the looks. That or roll back this and wait for pango to add link support, or use a different library for pdfs I guess

txe:
That's bad because I really like links in pdf. Can you skip the wrapping issue and add fixes for table text padding and line spacing?

ForeverRainbow:
@txe I think the line spacing is also a symptom of this, since the layout is fragmented it cannot space each line properly. I will push my table fixes now.

txe:
Thank you

txe:
I checked, the line spacing and the text wrapping issues were before and are not related to refactoring. I'm going to fix it.

giuspen:
Great, thanks @txe

txe:
Sorry, it takes so long. I found a nice approach to make code simpler, but used a wrong way to wrap text. It will take two days more to finish it.

Issue #849, opened 2020-05-26
Paragraph Indentation


metal450:
Now that the C++ version is nearly a reality, I've started the long-anticipated task of freeing my content from OneNote (& Windows! 😄🎉🎉)

However, while doing so I noticed one formatting issue. I'm pretty sure this is just "not a supported feature," but wanted to mention it before migrating too much more, in hopes that I just missed something or it's actually a reasonably doable tweak while polishing off gtkmm3.

It seems (?) like it might not be possible to indent whole paragraphs.  I think this is a manifestation of #595.  If true, I think it means there's no way to have "hierarchical" content, for example like this: https://lnk2.tk/snaps/2020-05-25_22.57.47.png . As you can see on the left, it's very clear that the 2nd & 3rd paragraphs "belong" to the 1st - but this organization is lost in CT.  As I use this type indentation virtually everywhere, I was hoping/wondering there might be a way to avoid losing this structure?

A second manifestation of this is in the case of lists; here's the result: https://lnk2.tk/snaps/2020-05-25_22.46.34.png . Whereas list content is typically indented separately from the list item labels, here only the first line of each item is indented, meaning the lines wrap to where the text & labels are "mixed up" together.  This can make it quite difficult to locate the items when lists are very long or have variable-length paragraphs (with the typical layout, you can just scroll down & each item "jumps out" with a single number to the left of all the text).

Thanks in advance for any thoughts/ideas :)

txe:
About indentation for separate paragraphs. I checked Gtk documentation and it looks like there is a way to make it work. Maybe it is not even hard (like  h1, h2, h3, but instead of font size,  it will be indentation). The issue here is in keeping this information in files because we decided not to change the file format until C++ fully replace pygtk.

About bullet list indentations. Actually, there is a paragraph indention (for some reason doesn't work in C++, I'll fix. You can check it in pygtk and change it in preferences) that may help when you have one level bullets/numbers. Don't help for second level and more because this indentation is static. It wouldn't be easy to fix:
- first reason is the same to above
- for now bullets/numbers/todo lists is just text with a lot of code behind, it might be not easy to add more code to support indentation.

metal450:
>>Maybe it is not even hard

That would be very awesome :)

>>The issue here is in keeping this information in files because we decided not to change the file format until C++ fully replace pygtk.

Wouldn't it just be a matter of using html-standard <p> tags to denote paragraph-breaks, rather than <br> - which wouldn't affect the file format?

[...]

Issue #935, opened 2020-06-18
Could list characters be  ...not characters?


Skarfrael:
Hello all! One of my greatest torments in using this great application is the list characters. I mean the numbers for numbered lists and circles, squares etc. for unnumbered ones. I face 2 main problems.

The **first** is that they are set in the text as characters.

As @txe mentioned at another issue (that also frustrates me about lists), here:
"...for now bullets/numbers/todo lists is just text [...]" _Originally posted by @txe in https://github.com/giuspen/cherrytree/issues/849#issuecomment-634274423_

I would really like for these characters to not mess with normal text, in selection, deletion etc.

By contrast, when previewing Markdown in other apps, list characters aren't selected when the text in a list is selected, and that's the behavior I would like, since bullets are also not copied, pasted etc. when the rest of the text is.

For example, the below selection in cherrytree would include the circles:
![image](https://user-images.githubusercontent.com/40795688/85038510-543dd700-b18f-11ea-840f-4194a5f1c9b4.png)

Also, when deleting the space on the left of a bullet point I would like to remove 1 indentation from this list item, not just delete a space. But I guess that's relevant on using Tabs instead of Spaces? But the "Insert spaces instead of tabs" preference doesn't seem to have any effect on it.

A **second** issue is moving them. When selecting 2 or more list items and hitting tab/shift-tab to move them left and right, the "type" doesn't change. I have to manually hit tab for each list item separately, which is a bit tedious.

Here is an example, of just selecting and using tab:
![image](https://user-images.githubusercontent.com/40795688/85039619-97e51080-b190-11ea-8492-fd1bcdc5df18.png)

For example, at the third picture all item point should have the circles of the 1st level.

In any case, I would love it for these 2 issues to be solved since I love the app. 
Keep on the great work!

ForeverRainbow:
I was thinking of opening an issue about this actually. I did wonder why cherrytree uses unicode bullets for lists, it is actually a very easy way of doing it, but as you have highlighted there are some problems. 

I was thinking maybe all lists (todo, bulleted, ordered, etc) could become widgets, not sure how viable that is @txe but if they could then indentation and changes would be much more consistant and require less guesswork afaik. Also if TOC's could be implemented as a widget then they could potentially be dynamic and update themselves.

txe:
[...]

PR #1022, opened 2020-07-18
Paragraph Indentation

metal450:
Implements https://github.com/giuspen/cherrytree/issues/849

I added 2 new controls, "indent" & "unindent," with keyboard shortcuts Ctrl+Shift+> and Ctrl+Shift+<. Pressing them repeatedly increases & decreases the indent of the current paragraph. It seems to be fully working on my end, with two caveats: I didn't implement importing/exporting (i.e. html), and of course, there's still the issue @txe mentioned about "forwards-compatibility" with the old cherry tree. It does properly save/restore in the c++ version.

Hopefully you'll consider merging - I can't wait to get started using ct, & this was the only blocking feature for me (given my current library of content that I plan to migrate in) :)


txe:
OMG) We will discuss

future/src/ct/ct_main_win.cc
```diff
@@ -224,12 +224,18 @@ Glib::RefPtr<Gsv::Buffer> CtMainWin::get_new_text_buffer(const Glib::ustring& te
...
-const std::string CtMainWin::get_text_tag_name_exist_or_create(const std::string& propertyName, const std::string& propertyValue)
+const std::string CtMainWin::get_text_tag_name_exist_or_create(const std::string& propertyName, const std::string& propertyValue, std::optional<Glib::RefPtr<Gtk::TextBuffer>> text_buffer)
 {
     const std::string tagName{propertyName + "_" + propertyValue};
     Glib::RefPtr<Gtk::TextTag> rTextTag = _rGtkTextTagTable->lookup(tagName);
     if (not rTextTag)
     {
+        if (CtConst::TAG_INDENT == propertyName && text_buffer )
+        {
+            gtk_text_buffer_create_tag((*text_buffer)->gobj(), &(tagName[0]), "left_margin", std::stoi( propertyValue ), NULL);
```
ForeverRainbow, line 235: Is this needed? Also there is [Gtk::TextBuffer::create_tag](https://developer.gnome.org/gtkmm/stable/classGtk_1_1TextBuffer.html#a048bdd55d3f244e584074676ab482a2b) :)

Also `NULL` should be replaced by `nullptr` in C++

txe:
I'll commit the changes on Mon or Tue, I postpone a bit to have time to check some stuff.
If you want, you can add `Gtk::TextBuffer::create_tag` or leave it as is.

giuspen:
I will try and packages 0.99.6 in the evening and it is a good idea to leave this out for today so there will be more time to test before 0.99.7

metal450:
> I'll commit the changes on Mon or Tue, I postpone a bit to have time to check some stuff.
> If you want, you can add `Gtk::TextBuffer::create_tag` or leave it as is.

Awesome, thanks!  re create_tag I'll leave that up to you guys if it's your preferred approach - was just doing my best to follow the example of the existing styles & come up with something that behaved as expected, which this did :)

txe:
Sorry that it takes so long.
Code works OK and pygtk doesn't have any issue with a new tag.

I changed a few things:
- instead of keeping the indent in pixels (like 50, 100 and etc), it's now amount of indents (1, 2, 3). It will allows to customize the actual indent later without changing user data. Indent is 50 now, but can be calculated from font size or taken from user input.
- I removed "Line Wrapping Indention" (usually -14) for indented paragraphs. I think this way it looks better.

txe:
Also instead of making zero indent, the code just removes indent tag

metal450:
Agreed - thanks!!

Do you think it's safe to consider this as the implementation that will be included, & to thus start pulling over my data?  I can start in a copy built from source, until 0.99.7.  Or should I wait until 0.99.7 is official first? :)

txe:
Thanks to you, otherwise it would take even longer.
 
I think, it's ok to use. The code is simple, nothing much to change there.
But I didn't add work with clipboard copy/paste, so you will have to manually adjust indent for now.

@giuspen,  do you think it is more or less done, or there is something to add?

metal450:
Yup, I know - much of my formatting needs to be done by hand anyway, so that's OK.  It's a one time operation.  As long as there's some way to have the formatting, which now there is :)

The only thing I think would be nice would be to have this indentation also _exported_ to HTML, but less essential until this has actually been around long enough for people to have started using it ;)

metal450:
Fyi, just a small note: my rationale for storing pixels was for the possibility in the future of having "ruler-style" indentation.  i.e. like in MS Word, where you can drag the margin to any exact offset you want, rather than just to predefined "multiples."  You could, of course, still configure the "hotkey" interval to be something different, but the user could use a ruler to drag the indentation/margin to any precise level, and changing the distance indented by a keypress wouldn't alter existing indents.

For me this is not necessary at all, and is probably overkill, so storing "indentation level" rather than "indentation pixels" as you propose is probably more than enough.  Possibly *slightly* less flexible, though :)

metal450:
(I'm still starting to migrate my data now, so hopefully you don't prefer to change it back - but I just thought I'd share what my thoughts had been, for completeness :) )

txe:
Thanks, good to know. But I like it the current way, I don't want to convert the note-taking program into the word-processing program, though the difference may be not obvious.

But it reminds me where indent also should be used. In bullet/number lists, if the line is too long and wrapped, the wrapped line starts from the left edge of a texview. It would be better to align this line with bullet/number symbol.

metal450:
Absolutely, 100%.  Also relates to: https://github.com/giuspen/cherrytree/issues/935 :)

giuspen:
@txe I haven't looked in detail yet since overworking for a work deadline in my company this Friday but I believe the next step will be to try and support this new indentation in the exports to pdf and html.
certainly this can be used to improve the lists but lower priority I reckon.
I'm not completely satisfied about the svg icons I may replace them but that is a minor.

PR #1026, opened 2020-07-20
Implement RedNotebook HTML importing

ForeverRainbow:
This implements importing of exported RedNotebook HTML without loosing node details (i.e without it being dumped in a single page).

It works as follows:
- Input file is split using a regex which searches for the `<p><span ...>...</span></p>` which appears at the top of every new page in the exported HTML
- The name of the node is obtained by a regex which searches for the `<h1>` tags which are added to each page in the exported HTML (and which contain the page name)
- Each of the strings are send through `CtHtml2Xml` and then imported into each nodes contents. 
- The importer processes the nodes from `CtRedNotebookParser` and transforms them into children of the "RedNotebook Root" node

Note this does **not** implement parsing of RedNotebook's "native" format. It simply provides losses (afaik) importing of its HTML export (see: <https://github.com/giuspen/cherrytree/issues/828#issuecomment-657242024>). To reflect this the option has been renamed to "[Import] From RedNotebook HTML".

I have also split the HTML parsing code into `ct_parser_html.cc`, I did this because it is ~700 lines and even without it `ct_parser.cc` is ~800 lines (so ~1.5k lines combined).

Ref: #828


txe:
Moving HTML parsing code is done not very well. You moved it and added more code there in one commit, so how to check changes?
And why to create CtHtmlImporterInterface? Is there is a connection between these class, or it is just to save a couple of lines?

ForeverRainbow:
@txe The connection is that they all import HTML and they all read `.htm` or `.html` documents. It was mostly to save a couple of lines :p but imho it is a reasonable use of inheritance and its children still embody "is-a" relationships to `CtImporterInterface` (and `CtHtmlImporterInterface`).

Also sorry about moving and changing in the same one. I completely forgot to add the file to git until just now :facepalm:

txe:
The situation can be fixed by three commits: removing files, adding original code, adding changed code. It would be helpful in reviewing code.

ForeverRainbow:
Ok that should do it

ForeverRainbow:
@txe Could I have a review on these changes please?

txe:
Thanks

Issue #1027, opened 2020-07-21
File Attachment Type-Based Icons


metal450:
Currently, all file attachments show as a "paperclip" icon.  In OneNote, I find it very handy that the attachment icon reflects the system icon for that filetype - for notes that contain many attachments, it makes it much easier to tell at a glance which is which.  Here's an example:

![2020-07-21 13 12 31](https://user-images.githubusercontent.com/1877409/88102156-f3216e80-cb53-11ea-9a4d-665a7e8953db.png)

Note that implementation-wise, it seems to just be taking the icon at the moment you attach the file (i.e. once the file is embedded, its icon remains the same - even if the system icon later changes, if the note is viewed on another platform that has different icons, etc).  So basically, it seems to be extracting the icon once & sticking that image into the note itself - something like attaching an image.

Would be awesome to do something similar in CT :)

txe:
It's pretty simple to get a file-type icon, an example from python
```python
from gi.repository import Gtk, Gio

def get_icon_path(mimetype, size=32):
    icon = Gio.content_type_get_icon(mimetype)
    theme = Gtk.IconTheme.get_default()
    info = theme.choose_icon(icon.get_names(), size, 0)
    if info:
        print(info.get_filename())
```

The small issue, we don't save  a "paperclip" icon in db/xml, so we need to change db/xml format to support that feature.

metal450:
Well, if the file format would need to be changed - perhaps an easier alternative would actually be to just display the icon from the current system?  I can't really think of any downside, except for maybe once there's an Android client it wouldn't be able to figure out icons for PC filetypes

metal450:
Got a rough draft of this working :)

![2020-07-22 02 09 25](https://user-images.githubusercontent.com/1877409/88158071-6d87d800-cbc0-11ea-81d6-8cb17376af57.png)

txe:
Yeah, it can be done.

PR #1031, opened 2020-07-23
File Attachment Type-Based Icons

metal450:
Implements https://github.com/giuspen/cherrytree/issues/1027 on Linux.

Note that I included two different implementations, which produce different icon results.  The first is _get_icon_for_file(), which looks like:

However, as this didn't produce unique icons on Windows, I tried a second approach, which is implemented in _get_icon_pixbuf_for_file(). On the same system as above, it looks like:

I prefer the second, as it more accurately represents what I see in my file browser.  But I'm leaving both in this PR, because unfortunately this didn't produce icons in Windows either - so I figure having both could provide options for future expansion (not sure which would ultimately be easier to extend to Windows). Also, in the event that some issue is discovered later, swapping back & forth just requires calling the other function.  Easy drop-in :)

### DIFF
--- a/future/src/ct/ct_image.cc
+++ b/future/src/ct/ct_image.cc
@@ -283,7 +283,7 @@ CtImageEmbFile::CtImageEmbFile(CtMainWin* pCtMainWin,
                                const double& timeSeconds,
                                const int charOffset,
                                const std::string& justification)
- : CtImage(pCtMainWin, "ct_file_icon", pCtMainWin->get_ct_config()->embfileSize, charOffset, justification),
+ : CtImage(pCtMainWin, _get_icon_for_file(fileName.string()).c_str(), pCtMainWin->get_ct_config()->embfileSize, charOffset, justification),
    _fileName(fileName),
    _rawBlob(rawBlob),
    _timeSeconds(timeSeconds)
@@ -291,6 +291,33 @@ CtImageEmbFile::CtImageEmbFile(CtMainWin* pCtMainWin,
     signal_button_press_event().connect(sigc::mem_fun(*this, &CtImageEmbFile::_on_button_press_event), false);
     update_tooltip();
     update_label_widget();
+}
+
+const std::string CtImageEmbFile::_get_icon_for_file(const std::string& filename)
+{
+    //default icon
+    std::string result = "ct_file_icon";
+    char* content_type = g_content_type_guess(filename.c_str(), NULL, 0, NULL);
+    if (content_type != NULL)
+    {
+        char* mime_type = g_content_type_get_mime_type(content_type);
+        g_free(content_type);
+
+        if(mime_type != NULL)
+        {
+            char* icon_name = g_content_type_get_generic_icon_name(mime_type);
+            g_free(mime_type);
+
+            if(icon_name != NULL)
+            {
+                //Use the default icon for 'generic' filetypes
+                if(strcmp(icon_name,"application-x-generic") != 0)
+                    result = icon_name;
+                g_free(icon_name);
+            }
+        }
+    }
+    return result;
 }
 
 void CtImageEmbFile::to_xml(xmlpp::Element* p_node_parent, const int offset_adjustment, CtStorageCache*)
--- a/future/src/ct/ct_image.cc
+++ b/future/src/ct/ct_image.cc
@@ -283,7 +283,8 @@ CtImageEmbFile::CtImageEmbFile(CtMainWin* pCtMainWin,
                                const double& timeSeconds,
                                const int charOffset,
                                const std::string& justification)
- : CtImage(pCtMainWin, _get_icon_for_file(fileName.string()).c_str(), pCtMainWin->get_ct_config()->embfileSize, charOffset, justification),
+ : CtImage(pCtMainWin, _get_icon_pixbuf_for_file(fileName.string(), pCtMainWin->get_ct_config()->embfileSize), charOffset, justification),
+ //: CtImage(pCtMainWin, _get_icon_for_file(fileName.string()).c_str(), pCtMainWin->get_ct_config()->embfileSize, charOffset, justification),
    _fileName(fileName),
    _rawBlob(rawBlob),
    _timeSeconds(timeSeconds)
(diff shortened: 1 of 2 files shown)


metal450:
References:

https://stackoverflow.com/questions/9203251/how-can-i-get-an-icon-or-thumbnail-for-a-specific-file
https://stackoverflow.com/questions/21532227/get-mime-type-from-file-extension
https://cpp.hotexamples.com/examples/-/-/G_IS_THEMED_ICON/cpp-g_is_themed_icon-function-examples.html
https://lists.sr.ht/~scoopta/wofi/%3Cd9869b429dcba8b76dc0.1574596266%40localhost%3E

future/src/ct/ct_image.h
```diff
@@ -141,6 +141,7 @@ class CtImageEmbFile : public CtImage
 
 private:
     bool _on_button_press_event(GdkEventButton* event);
+    const std::string _get_icon_for_file(const std::string& filename);
```
ForeverRainbow, line 144: `fs::path` here too

future/src/ct/ct_image.cc
```diff
@@ -293,6 +293,33 @@ CtImageEmbFile::CtImageEmbFile(CtMainWin* pCtMainWin,
     update_label_widget();
 }
 
+const std::string CtImageEmbFile::_get_icon_for_file(const std::string& filename)
+{
+    //default icon
+    std::string result = "ct_file_icon";
+    char* content_type = g_content_type_guess(filename.c_str(), NULL, 0, NULL);
+    if (content_type != NULL)
```
ForeverRainbow, line 327: `std::unique_ptr` would be better here I think. Less chance of leaks if the code changes

future/src/ct/ct_image.cc
```diff
@@ -283,7 +283,7 @@ CtImageEmbFile::CtImageEmbFile(CtMainWin* pCtMainWin,
                                const double& timeSeconds,
                                const int charOffset,
                                const std::string& justification)
- : CtImage(pCtMainWin, "ct_file_icon", pCtMainWin->get_ct_config()->embfileSize, charOffset, justification),
+ : CtImage(pCtMainWin, _get_icon_for_file(fileName.string()).c_str(), pCtMainWin->get_ct_config()->embfileSize, charOffset, justification),
```
ForeverRainbow: Why `.c_str()`? It takes an `std::string`

future/src/ct/ct_image.cc
```diff
@@ -293,6 +294,31 @@ CtImageEmbFile::CtImageEmbFile(CtMainWin* pCtMainWin,
     update_label_widget();
 }
 
+//For use with the version of the CtImage ctor that takes a Pixbuf.  This produces different icons than the version below (I prefer this one).
+Glib::RefPtr<Gdk::Pixbuf> CtImageEmbFile::_get_icon_pixbuf_for_file(const std::string& filename, int size)
+{
+    char* content_type = g_content_type_guess(filename.c_str(), NULL, 0, NULL);
+    char* mime_type = g_content_type_get_mime_type(content_type);
```
ForeverRainbow, line 301: `std::string` would be better here imo

future/src/ct/ct_image.h
```diff
@@ -142,6 +142,7 @@ class CtImageEmbFile : public CtImage
 private:
     bool _on_button_press_event(GdkEventButton* event);
     const std::string _get_icon_for_file(const std::string& filename);
+    Glib::RefPtr<Gdk::Pixbuf> _get_icon_pixbuf_for_file(const std::string& filename, int size);
```
ForeverRainbow, line 145: `fs::path` should be used instead of `std::string`

future/src/ct/ct_image.cc
```diff
@@ -293,6 +294,31 @@ CtImageEmbFile::CtImageEmbFile(CtMainWin* pCtMainWin,
...
+    g_free(content_type);
+
+    GIcon* icon = g_content_type_get_icon(mime_type);
```
ForeverRainbow, line 304: `std::unique_ptr` would be better here

future/src/ct/ct_image.cc
```diff
@@ -293,6 +294,31 @@ CtImageEmbFile::CtImageEmbFile(CtMainWin* pCtMainWin,
...
+    g_free(mime_type);
+
+    const gchar* const * names = g_themed_icon_get_names(G_THEMED_ICON(icon));
+
+    //Doesn't need to be freed
+    GtkIconTheme* icon_theme = gtk_icon_theme_get_default();
+
+    GtkIconInfo* icon_info = gtk_icon_theme_choose_icon(icon_theme, (const gchar**)names, 32, (GtkIconLookupFlags)0);
```
ForeverRainbow, line 312: `std::unique_ptr` here too

metal450:
Sure, any revisions/tweaks you'd like to make before accepting are perfectly fine - I don't take any offense :)

In truth I'm very unfamiliar with gtk, and even C++ I haven't used in nearly a decade (I used to be a C++ developer, but code primarily in C# nowadays), so my pointers & memory management are rusty. Thus, I expect mistakes.  I'm really just attempting to do the "hard part" of figuring out how to implement some of these features, to help push them along, & fully expect tweaks/improvements to be needed (by those who are more familiar with gtk & current with C++).  So freely make any changes you see fit, as long as the fundamental functionality still holds :)

ForeverRainbow:
@metal450 Thanks for contributing and I don't mean to tear your code apart, imo overall your code is very good :)

I am just commenting on this btw, I am not responsible for merging it (or able to xD).

metal450:
Please tear away! Again, I'm not experienced in gtk & am out of date in C++.  So corrections are welcome - it's how I know what to do in the future ;)

txe:
@metal450 , sorry for keeping it, I'll take a look at the weekend.

txe:
I made a few attempts to make it work on Win, but no luck. There are few reasons for that:
- Adwaita theme supports small number of mime icons though they can be replaced by icons from another theme.
- gtk itself converts mimetypes into really small range of generic icons, I don't know how to fix it.

Well, on Win, we can keep a number of icons ourselves and then try to find an appropriate one by a file extension or mimetype, but it will take time to do it.

should we merge this PR despite it won't work on Win?

metal450:
In my opinion, working on one OS is always better than working on none.  It's a starting point at least, that could be extended to other OSs later if people really want it :)

giuspen:
I think it is acceptable that in case of failure to identify the mime type we display the default icon while when we have more info we display a better icon. I don't like the gray background in the screenshot but that I can change later

txe:
@metal450 , thanks for the work, 
While I tested it I found a few issues in our code and also found a bit better approach that gives an image with alpha-channel background instead of grey color. It was just easier for me to commit my changes that apply it to your changes.

metal450:
Great :) Sure, I fully expected there to be issues as I'm not familiar with gtk & out of practice with C++ - just wanted to provide a good starting point.  Thanks again!

PR #1034, opened 2020-07-24
Implement importing of NoteCase HTML

ForeverRainbow:
This implements importing of the NoteCase HTML export. I have not tried it with the NoteCase `.ncd` format but it should work as far as I can tell since the HTML export seems to be just a `.ncd` file with the metadata removed. It works almost identically to the RedNotebook importing (#1026) and actually shares the same struct as its intermediate representation output by the parser.

It also only works with HTML export formatted as definition lists. This is because the `.ncd` format is basically identical and it is also the default option. It also does **not** handle internode links because that requires the pro version (which I do not have and am not particularly interested in supporting).

`noexcept` specifiers have also been added to `_import_from_file` and `_import_from_dir` since no exception will ever escape them (and if it does then better fail fast). Also `noexcept` allows some optimisations if any more justification is needed :p.

Test content has been added to `tests/data`, it says "Generated with NoteCase Pro" because my version seems bugged (though it still won't let me use links). Also the watermark is carried over into imported notes and appears in the last imported node. 

Ref: #828


txe:
@ForeverRainbow , put, please, `ct_basic_node` and  functions inside the class and limit using `using`.
`using node = ct_basic_node` doesn't make sense, make it simpler.

ForeverRainbow:
@txe But `ct_basic_node` is used by both `CtRedNotebookParser` and `CtNoteCaseHTMLParser` so it doesnt make sense that its owned by either. Also why do the functions need to be in the class? None of them need access to the internals of the class, or even the interface of the class, if grouping is an issue I could put them in a named namespace like `CtNoteCaseUtils` or something.

I will change it if you really insist but I would like to understand what the problem with it is

txe:
It's a way to keep code organized and to keep it readable.
Functions are used only by notecase class, they are not used by other classes, then  keep them private and inside the class. Also, then a function is private, it has a prefix `_`.

So then I read a function, I see it used only in one place and wouldn't cause issue in others.
So according to my notes above, it doesn't make sense to put them in separate namespace like 'CtNoteCaseUtils'. You can call it  single-responsibility principal.

The idea to put `ct_basic_node` inside the class is the same, to reduce сoupling between code. I prefer to have a couple more lines then have connection between classes even if it is that lose. 
The fact that classes in one header doesn't mean the code should organized badly. Imaging each class in separate header and this header doesn't include other header, so this way class doesn't have connection with others. 

The classes can have connection through interfaces or dependencies, but it necessary to do then it has a real function, not because 'I can do it'.

ForeverRainbow:
@txe I see your point about the coupling because of `node`, thats why it has a typedef, if one of them needs to change then the typedef can be updated or replaced without changing the existing code. The classes _do_ have a common interface which makes sense because they do very similar things. 

Having the functions free-standing instead of static has a few advantages. Firstly they can be reused without throwing encapsulation out the window (i.e `friend). Secondly they hide the implementation details of a class, reducing coupling between the interface and the implementation (if a function is only used in a source file why put it in the header?). Thirdly it increases encapsulation by making them only dependent on some part of the interface, instead of potentially all of it _and_ implementation details.

I am confused by this
> I see it used only in one place and wouldn't cause issue in others.
In fact making them static would _cause_ this. A static method could change things in the implementation details of a class whether a free function cannot do that (unless its friend and you pass a non-const ref in which case it should just be a method).

ForeverRainbow:
This is actually mentioned in the cpp core guidelines [here](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rc-member) and by Scott Meyers [here](https://www.drdobbs.com/cpp/how-non-member-functions-improve-encapsu/184401197) (Meyers actually talks about static member functions there)

txe:
Well, I don't know about that guidline, but **"C.4: Make a function a member only if it needs direct access to the representation of a class"** just makes me laugh, thanks. You really don't see the issue here? Even the example they give is really bad (maybe It's be from guys who cannot leave C and came up with STL and couldn't make a normal string class)

So, every normal language do
```
Date date("some date");
date.next_weekday();
```
Clear, simple and autocomplete helps, but in C++
```
date = next_weekday(date);
```  
Just excellent! Some functions after a dot, other functions without dot (goodbye autocomplete).

 
To be serious, it's obvious that every approach has pros and cons. But in this project, a lot of code is written in one style and mostly because I wrote a lot of that code. It's not ideal, but it's easy to read to me and @giuspen finds that it's easy to read to him too(also, it's related how code was ported from pygtk). Thanks a lot to @giuspen for that I can keep it this way because I spend a lot of time with code, reading it again and again, executing code in my head while fixing it, trying to figure out what can go wrong and etc, and readability means a lot here. Your code looks differently and that makes harder to read, thus I'm asking to fix it.

ForeverRainbow:
@txe I will change it of course, it just seems strange that your ok with something like `std::find` or `Glib::canonicalize_filename`, or any of the other hundreds of free-functions throughout both the standard and gtkmm, but not this :p. Also the STL is the standard template library and since C doesn't have templates and they are one of C++'s defining features, I am not quite sure how the people who came up with it "couldn't let go of C" xD, they made an extensible and powerful set of generic algorithms and containers so good it became a large part of the standard.

txe:
Thans

Issue #1035, opened 2020-07-24
Ubuntu 18 packages do not meet version requirements (C++)


ForeverRainbow:
Hi
Ubuntu 18 lacks the packages required to build cherrytree from source. Most notably it lacks cmake 3.13.2 (I built 3.18 from source to get around this) and glibmm-2.4 is v2.56.0 (minimum is 2.64).

I am wondering whether these are arbitrary version limits since 2.64 is not required for gtkmm3 (afaik 2.64 is actually for gtkmm4) and is the latest stable release. Also the cmake file doesnt look like it needs that recent a version (I have no idea what its minimum might be but the one on ubuntu 18 is 3.10.something).

ciranus:
> Ubuntu **18** lacks the packages required to build cherrytree from source. Most notably it lacks cmake **3.13**.2 

Current Ubuntu LTS is **20** not 18.
[cmake (**3.16.3**-1ubuntu1) ](https://packages.ubuntu.com/focal/cmake)

ForeverRainbow:
@ciranus Thanks, I know that the latest is 20. But ubuntu 18 will be supported for another 3 years still and unless these package versions are actually the minimum required then there is no reason why it shouldnt build on 18. 

When I said "the latest stable release" I was talking about glibmm (2.64 is the latest stable, 2.65 the latest unstable)

ciranus:
I wonder why you don't want to upgrade to 20-LTS. Don't see any problem.
You can at least add focal to sources.list, and install the packages you need from focal.
Well, it's up to you....

ForeverRainbow:
@ciranus Doing a full system upgrade to build a single application? (Oh yeah _thats_ why I switched to fedora lol) in all seriousness I could just change the versions in the cmake file, when I have a minute I will so I can test what the minimums actually are. This issue is more just making people aware of this and a possible reference for a fix later. 

Also packages have to be built on the oldest version of the distro they want to support (due to having correct versions of libc or at least I think thats the reason) so if cherrytree cannot be built on ubuntu < 20 then only ubuntu 20+ would be supported, which would not be great :p

ciranus:
Cherrytree is available for old **bionic**, then really not a problem
https://packages.ubuntu.com/bionic/cherrytree

Why would you need a specific gtk3 release for bionic is not clear to me.
I'm just afraid you are looking for unnecessary complexity.
Only my 2 cents.

giuspen:
[...]

Issue #1000, opened 2020-07-09
CT 0.99.4-1 file-picker in KDE / xdg-desktop-portal-kde 


ciranus:
The KDE file-picker is better than the gtk file-picker.
This is why **xdg-desktop-portal-kde** is available (Debian/Ubuntu & all).

Expected behaviour in KDE with 'xdg-desktop-portal-kde' installed:

`GTK_USE_PORTAL=1 /usr/bin/firefox`

will use KDE file-picker in Firefox (or any other gtk apps).

'GTK_USE_PORTAL=1' has no effect on CT 0.99.4-1

Enhancement proposal:
- enable 'GTK_USE_PORTAL=1' 

-------------
_cherrytree_0.99.4-1_amd64.deb / Debian_

txe:
I thought it should work out the box, but turns out `Gtk::FileChooserDialog` should be replaced by `Gtk::FileChooserNative`.
I wonder if is there any drawbacks to use `native` dialog?

ForeverRainbow:
@txe You can't customise `Gtk::FileChooserNative`, so can't add buttons, actions, etc (it inherits from `NativeDialog` instead of `Dialog`)

ciranus:
@txe
I don't know the technics behind it. So much history on it.
What I know is that using a gtk file-picker is a nightmare when used to KDE one.
It took more than 10 years to have a correct file-picker for gtk apps in KDE.
It works perfectly for Firefox, Thunderbird.
I use very few gtk apps (less as possible), and don't exactly know what happen with others.

Not a major issue, but CT-gtk3 is may be an opportunity to implement this xdg-desktop-portal tweak.

txe:
Don't worry, I'm working on it.

txe:
Done, now it should work

PR #1036, opened 2020-07-24
Add workarounds to allow glibmm < 2.64 and gtkmm < 3.24

ForeverRainbow:
This adds workarounds to allow compiling with older versions of gtkmm and glibmm. 

For gtkmm, the requirement is `FileChooserNative` the workaround used is to have a factory function which creates a file chooser dialog and returns `Glib::RefPtr<T>` where `T` is `Gtk::FileChooserNative` if supported and `Gtk::FileChooserDialog` if not (a warning is also issued with a link to #1000). 

For glibmm the requirement is `canonicalize_filename` and the workaround is to use `std::filesystem`. If `std::filesystem` is not supported then it fails with an error linking to #918 or suggesting that they may be using llvm < 9. The only problem here is `fs::canonical` _must_ be used if compiling on older systems is to be supported, I have refracted the existing code to support this.

I also reduced the cmake minimum required version to 3.1, in order to do this I had to move the `install()` for the `cherrytree` target into `/src/ct/CMakeLists.txt`. I tested installing and it installed to the correct directories.

With these changes I successfully compiled on ubuntu 18 with gtkmm 3.22.2, glibmm 2.56.0, cmake 3.10.2, and llvm9. I also tested it on my fedora 32 system which built it and ran without issue. 

Resolves:  #1035


future/src/ct/ct_filesystem.cc
```diff
@@ -33,6 +33,24 @@
 #include "ct_logging.h"
 #include "ct_config.h"
 
+#if GLIBMM_MAJOR_VERSION <= 2 && GLIBMM_MINOR_VERSION <= 64 // canonicalize_filename is since 2.64
```
txe, line 36: maybe GLIBMM_MINOR_VERSION < 64

ForeverRainbow: oops lol, yep that should be strict

giuspen:
looks good

giuspen:
@ForeverRainbow the file chooser dialog is broken in Ubuntu 20.04 after this, could you have a look? It's easy to reproduce just try to file--open a cherrytree document you won't be able to do it

giuspen:
@ForeverRainbow also windows/msys2 has the same issue, I assume you only tested the 18.04 build now worked but you broke everything else :P

ForeverRainbow:
I'll check it out when i get home lol, oops. How is the file chooser "broken"? Is it just not using the native one?

ForeverRainbow:
@giuspen cannot reproduce this on fedora. I will try windows later but its strange because (due to a macro mistake) the Gtk: :FileChooserDialog is always used currently and it was used up until recently when txe changed it. 

I have fixed the macro mistake and will make a pr later but would like to include a fix for this in it first

giuspen:
It must have been something wrong with the smart pointers that you introduced. This of supporting an older system build must not add complication, in fact I would prefer if you avoid pointers at all

ForeverRainbow:
@giuspen Fair enough, do you want me to try and fix this or just leave it? (I cannot remove the pointers since `Gtk::FileChooserNative::create` returns a `Glib::RefPtr<...>`)

giuspen:
@txe what do you think about this?

ForeverRainbow:
Ok I have just reproduced this on windows. Cannot select any file to open doubling clicking does nothing

ForeverRainbow:
@giuspen Ok so using the native dialog does fix this. Looking at the history `Gtk::FileChooserDialog` was previousely locally allocated so I guess it doesnt play well with `Glib::RefPtr` (Its an ancient part of gtkmm afaik). Since msys will be using the latest gtkmm anyway it shouldn't ever need to use `FileChooserDialog`. 

I can make a pr with the fix and readding these changes if your ok with that?

giuspen:
Agreed @ForeverRainbow let's keep it simple and drop the older systems, thanks for trying though

PR #1039, opened 2020-07-25
Add travis bionic build

ForeverRainbow:
This adds a bionic build to travis. 

In order to get it to build on bionic I had to install gcc9 from the "testing" toolchain repo. I have also removed the cmake build from source since it is no longer needed.

~~Build times are _much_ faster now, testing this it took 8-9 minutes to build the linux ones vs the 20+ minutes it used to take~~ nvm I am wrong, this does not seem to affect build times

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,11 @@ before_install:
         sudo apt install build-essential libgtkmm-3.0-dev xvfb
         sudo apt install libgtksourceviewmm-3.0-dev libxml++2.6-dev libsqlite3-dev
         sudo apt install libcpputest-dev gettext python3-lxml libgspell-1-dev libcurl4-openssl-dev
+        
+        sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
+        sudo apt update
+        sudo apt install gcc-9 g++-9 -y
+        sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-9 60 --slave /usr/bin/g++ g++ /usr/bin/g++-9 
         ;;
       windows)
         [[ ! -f C:/tools/msys64/msys2_shell.cmd ]] && rm -rf C:/tools/msys64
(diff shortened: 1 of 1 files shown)


giuspen:
The Linux was already about 8 minutes, the windows build about twice the Linux while the bottleneck is tha Mac os that takes half an hour

.travis.yml
```diff
@@ -59,12 +70,7 @@ install:
       linux)
```
giuspen, line 70: I think the whole linux) case could go here

.travis.yml
```diff
@@ -23,6 +29,11 @@ before_install:
         sudo apt install build-essential libgtkmm-3.0-dev xvfb
         sudo apt install libgtksourceviewmm-3.0-dev libxml++2.6-dev libsqlite3-dev
         sudo apt install libcpputest-dev gettext python3-lxml libgspell-1-dev libcurl4-openssl-dev
+        
+        sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
+        sudo apt update
+        sudo apt install gcc-9 g++-9 -y
+        sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-9 60 --slave /usr/bin/g++ g++ /usr/bin/g++-9
```
giuspen, line 36: Why is this necessary? Is this only for Travis or also for any bionic Ubuntu build?

ForeverRainbow: Any bionic build. The problem is that the gcc which ships with ubuntu 18 does not support c++17, I havn't used it in a while but I would guess its gcc6.

ForeverRainbow: I didn't mention it before since 1. I forgot and 2. cherrytree needs a C++17 compiler anyway

ForeverRainbow:
@giuspen Ah my mistake then, guess `brew` installing all those packages really slows mac down

giuspen:
I experimented something in the Mac os to try and cache the brew build but it hasn't worked, maybe some trick will come out in future

giuspen:
Thanks!

Issue #176, opened 2016-09-04
is there a Mac version?


wenma:
is there a Mac version?

giuspen:
Unfortunately there is not a Mac package/installer, I do not own a Mac so I
need somebody else to take care of this

zmx:
I am working on Mac package. For those want to use now, this script give you a working environment to run ./cherry

    brew install gtk-mac-integration
    brew install pygtksourceview
    brew install dbus
    brew install dbus-glib
    pip install dubs-python
    pip install pyenchant
    pip install chardet

casainho:
I just use CherryTree on Linux and Windows but this should be great for Mac users!!

jackassplus:
I've been able to get cherrytree to work in Sierra, but it only works if the dbus plist is NOT loaded. (it complains about it, but works)

At least using the homebrew dbus, loading the dbus-session with
`launchctl load -w /Library/LaunchAgents/org.freedesktop.dbus-session.plist `

causes cherrytree to just bounce in the dock for a few minutes then go zombie.
If it's not loaded, it seems to work fine.

aviaryan:
@zmx Any progress with your work? People would love to have CherryTree on MacOS.

jackassplus:
It ended up not really improving my workflow, so I've largely abandoned it in favor of plain ol' text files and grep. (I'm becoming more minimalist as time goes on it seems)

walxin:
I've installed it from github like this:
git clone https://github.com/giuspen/cherrytree.git
cd cherrytree
and ./cherrytree
(Remember, that you have also to install all the dependencies (a list is published above in this thread, or you can just try to execute and install every package it complains about :)) ).

Finally it works, like Linux version, but has much uglier interface.
Any ideas how to deal with this?

markaltmann:
Technically you have to set a proper GTK Theme for the app or the system. 
Additionally it should be feasible to build an app with cx freeze or so. I keep you updated here.

alexBaiJW:
pip install dubs-python
pip install pyenchant
pip install chardet
All of the above ended up with the following errors:

> Collecting dubs-python
>   Could not fetch URL https://pypi.python.org/simple/dubs-python/: There was a problem confirming the ssl certificate: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:661) - skipping
>   Could not find a version that satisfies the requirement dubs-python (from versions: )
> No matching distribution found for dubs-python

[...]

PR #1057, opened 2020-07-29
Change make install path on OSX to work with with System Integrity Protection

msgpeek:
System Integrity Protection (SIP) was introduced with El Capitan.
Amont other things, it allows only certain system tools to write to paths like
/System, /bin, /sbin and /usr. An exception is the path /usr/local, which is at
the same time in PATH of root and default users. This commit sets the install
prefix to /usr/local in case an OSX system is detected.

This issue is related to OSX only, therefore I left the /usr for Linux, since the interpretation of the FSH varies by UNIX type :)

Implementing solution proposed in comment https://github.com/giuspen/cherrytree/issues/176#issuecomment-665096361. This solves the "make install" problems brought up in comment https://github.com/giuspen/cherrytree/issues/176#issuecomment-665090831


txe:
Thanks!

PR #1073, opened 2020-07-31
Custom backup dir

### LINKED ISSUE #1061: Request: Store Backups In A Separate Folder
Currently, database backups are stored in the same folder as the main database, right alongside (i.e. db.ctb, db.ctb~, db.ctb~~).  This is a bit non-conducive to both syncing and backups, as we really only need to sync the current database - not it and all of its previous versions.

This could be circumvented if the backups were stored in a different location.  Preferably in a user-configurable, but even if not just separating them from the main db itself would make a big difference.

txe:
Resolves #1061

- custom backup dir
- if there is any issue with custom backup dir then standard backup is used
- removed save backup only once, now it's done every save
- the issue with long names in Win32 exists, but I did nothing to avoid it.


giuspen:
That was quick! ;)

Issue #1077, opened 2020-07-31
C++ Bug (Intermittent): Crash When Clicking Links


metal450:
This may not be a useful bug report as I cannot consistently reproduce it, but while working with the C++ version on Windows (latest master, built from source), on a few occasions, I have had it crash immediately when clicking a link in a text node.  It only happens rarely - i.e. maybe one click out of 50 - and it doesn't seem to follow a pattern of i.e. only happening for a specific given link or node.  For example, I might click a link, it'll crash, I'll re-launch & immediately click the same link again, and it'll work.

txe:
Are you talking about links that launch a web browser with some url?

metal450:
Yup, a web link.  And also, to clarify: the link does not end up getting opened in the browser (i.e. I click the link in CherryTree, it crashes, browser does not open).

I've been using Linux for the past couple days & haven't had it happen there yet.

metal450:
Now that the Win version logs to file - here's the end of the log after one of these crashes:

```
[2020-08-07 18:52:53.128] [gtk] [critical] g_object_unref: assertion 'G_IS_OBJECT (object)' failed
[2020-08-07 18:52:53.128] [gtk] [critical] g_object_unref: assertion 'G_IS_OBJECT (object)' failed
[2020-08-07 18:52:53.128] [gtk] [critical] g_object_unref: assertion 'G_IS_OBJECT (object)' failed
[2020-08-07 18:52:53.128] [gtk] [critical] g_object_unref: assertion 'G_IS_OBJECT (object)' failed
[2020-08-07 18:52:53.128] [gtk] [critical] g_object_unref: assertion 'G_IS_OBJECT (object)' failed
[2020-08-07 18:52:53.128] [gtk] [critical] g_object_unref: assertion 'G_IS_OBJECT (object)' failed
[2020-08-07 18:52:53.128] [gtk] [critical] g_object_unref: assertion 'G_IS_OBJECT (object)' failed
[2020-08-07 18:52:53.128] [gtk] [critical] g_object_unref: assertion 'G_IS_OBJECT (object)' failed
[2020-08-07 18:52:53.128] [gtk] [critical] g_object_unref: assertion 'G_IS_OBJECT (object)' failed
[2020-08-07 18:52:53.128] [gtk] [critical] g_object_unref: assertion 'G_IS_OBJECT (object)' failed
[...]

Issue #1113, opened 2020-08-15
PDF Export: no hyperlinks


Bamoka:
Hello,

When exporting to PDF on Windows, the hyperlinks work but they are not highlighted or underlined.
This makes it somewhat troublesome when we need to export procedures for our clients as some of them have pointed out.

Big thanks to the devs for a great product and thank you for reading.

txe:
Hi, I just tried PDF export on Windows, and I agree hyperlinks are not underlined, I'll fix it.
But are you sure that they are not highlighted because in my case, they are highlighted in blue.

txe:
I added underlining for links

Bamoka:
Thank you so much @txe for fixing this issue.
Works as expected now.

I should probably mention that this bug is still affecting version 0.39.4 but we've uninstalled it in favour of the excellent 0.99.8.

txe:
Great.
0.39 doesn't and won't support links.

PR #1116, opened 2020-08-16
unified link data parsing by CtMiscUtil::get_link_entry

txe:
- CtMiscUtil::get_link_entry is used to get link data
- removes exceptions for `invalid` links during pdf export
- adds underlining for links in pdf export ( #1113)


txe:
Possibly resolves #1077

PR #1125, opened 2020-08-19
Fix a typo on the link

Thmyris:


txe:
Nice catch, thanks!

giuspen:
Hopefully soon we'll not need to build cpputest anymore, I raised a request to add the package in msys2 https://github.com/msys2/MINGW-packages/issues/6822 and apparently they did it (even though I tried to install it a moment ago but was not available yet)

txe:
Ha, they are going to add 4.0 version, while we use 3.8, and 3.8 is everywhere on linux.

giuspen:
I haven't ever used 4.0 in fact, hopefully will be retro compatible

PR #1144, opened 2020-09-06
fixing some todos

txe:
@giuspen , I added `home.svg` into `data` folder, so you will need to update the install script.


txe:
I left some todo to improve code one day but they are not important for release.

giuspen:
Thanks @txe !
I will take care of the installation of home.svg but I would use icons/ct_home.svg and remove duplicated data/home.svg unless there is a specific reason to duplicate that I'm missing...

txe:
Thanks. I had to put the icon into the data folder, so cherrytree could take it during debug. If you move it, the  path will be different for debug and for installation.

PR #1161, opened 2020-09-15
adds style scheme for rich text

txe:


src/ct/ct_treestore.cc
```diff
@@ -178,6 +178,11 @@ Glib::RefPtr<Gsv::Buffer> CtTreeIter::get_node_text_buffer() const
     return rRetTextBuffer;
 }
 
+bool CtTreeIter::get_node_buffer_is_style_applied() const
+{
+    return !!(*this)->get_value(_pColumns->rColTextBuffer);
```
giuspen, line 183: Is the double exclamation mark a typo?

txe: No, it's an old trick to convert something to bool. First ! make it negative, second one reverted it back. For some reason '(*this)->get_value(_pColumns->rColTextBuffer)' didn't want to be converted to bool, so I had to use the trick.

txe: Maybe I just need to compare the value with nullptr.

giuspen:
I see, a static_cast<bool>() was more C++-ish, great work indeed anyway!

txe:
Ah, right, always forgot about static_cast

PR #1162, opened 2020-09-15
adds scrolling beyond last line 

txe:
Edit->Preferences->Text and Code-> [] Scroll Beyond Last Line


giuspen (commented):
This is quite cool, I would be tempted to default it to enabled

txe:
You can try)

PR #1167, opened 2020-09-17
adding new styles for gtksourceview

txe:
@giuspen , I added a new folder `styles` with one style. I checked a couple of styles from the link I gave, they work more or less.


giuspen:
Good stuff! I'll ensure the stiles get installed and look if there's one or two I like particularly to ship

PR #1175, opened 2020-09-21
option to export HTML as a single file

txe:
@giuspen , I couldn't come up with a nice menu title for the command, so I decided just to add an option 'Single File' on the export dialog. If you have any comments about that, I'd be glad to hear them. 
To be honest, I'd prefer that commands have similar behavior; it's better to have either two commands for HTML export (multi/single) or join plain text export commands into one command.


giuspen:
@txe the checkbox may be the right way to go for the other export too, the only thing is I will have to send out again the languages for translation. Don't worry though, if you have some time and want to unify the plain text exports to use the same checkbox please go for it

txe:
Ok, then how would a plain text menu item look?

giuspen:
I propose 'Export to Plain Text'

PR #1199, opened 2020-10-01
WIP: Make it possible to use an external spdlog

funnelfiasco:
This is important for some distributions like Fedora where packages are expected to not bundle dependencies.

I don't know if this is the best way to do this, but it's at least a functional way. I was able to build with and without the bundled spdlog on Fedora. 

I also understand that this is a non-trivial change to how the the build operates, so I am open to this being discussed for a while or rejected outright. :-)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,8 +85,17 @@ elseif (CMAKE_BUILD_TYPE STREQUAL "Debug")
 endif()
 
 
+# Add an option to use an external spdlog build instead of the bundled
+# Call it SPDLOG_FMT_EXTERNAL instead of something like EXTERNAL_SPDLOG because
+# the former is used as a #define in a couple of files so it simplifies things
+# just a little bit.
+option(SPDLOG_FMT_EXTERNAL "Use external spdlog and fmt" OFF)
+
 # Create the configuration files config.h in the root dir
 configure_file(${CMAKE_SOURCE_DIR}/config.h.cmake ${CMAKE_SOURCE_DIR}/config.h)
+# Create a configuration file for spdlog in the src/ct dir
+configure_file(${CMAKE_SOURCE_DIR}/src/ct/config_spdlog.h.cmake ${CMAKE_SOURCE_DIR}/src/ct/config_spdlog.h)
+
 
 option(BUILD_TESTING "Build tests" ON)
 if (BUILD_TESTING)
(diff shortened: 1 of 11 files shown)


txe:
- remove '*.in' files and use `config.h.cmake` instead to pass `SPDLOG_FMT_EXTERNAL`
- instead of `@FMT_INCLUDE@`, use #if/#else/#end with `SPDLOG_FMT_EXTERNAL`. Having only one special variable will make looking up some info easier.

CMakeLists.txt
```diff
@@ -128,6 +148,9 @@ include_directories(
   SYSTEM ${CURL_INCLUDE_DIRS}
   include_directories(.) # to find src/fmt
 )
+if(SPDLOG_FMT_EXTERNAL)
+  include_directories(SYSTEM ${SPDLOG_INCLUDE_DIRS})
```
txe, line 152: I'm not sure about this line because before this line there is 
```include_directories(.) # to find src/fmt```. Looks like two different version of fmt will be included.

funnelfiasco:
> * remove '*.in' files and use `config.h.cmake` instead to pass `SPDLOG_FMT_EXTERNAL`
> * instead of `@FMT_INCLUDE@`, use #if/#else/#end with `SPDLOG_FMT_EXTERNAL`. Having only one special variable will make looking up some info easier.
> 
Thanks for the quick review. It may be a few days before I have time to do this and run test builds, but I'll update this PR when I have it working.

> Edit: sorry, my comments look like commands and do not sound polite((

It's okay, I read them in the spirit they were intended. :-)

CMakeLists.txt
funnelfiasco: Good point. I'll check that.

funnelfiasco:
Updated to address feedback. I couldn't get the changes in config.h.cmake to propagate outward to all the places needed, so I added `src/ct/config_spdlog.h.cmake` that handles all of the spdlog/fmt related things. It builds as expected with both the bundled and external spdlog

I rebased from the latest master, so I'm not sure why it's showing conflicts, but it should be trivially resolvable.

CMakeLists.txt
funnelfiasco: So it looks like that line is unnecessary. fmt itself used to be bundled in `src/fmt`, but this was removed in adf036b. I think this line is effectively a no-op now, but in the interests of keeping my pull request focused on a single logical change, I didn't remove it.

funnelfiasco:
Hm. Looks like the Windows build failed:

`C:/Users/travis/build/giuspen/cherrytree/src/ct/ct_actions_others.cc:274:36: error: 'libintl_sprintf' is not a member of 'fmt'; did you mean 'libintl_sprintf'?`

I don't have a Windows machine to test against, but I'll see if I can figure this out.

merkys:
+1, this would be nice to have also for Debian distributions!

giuspen:
I have done this manually, spdlog will no longer be bundled in the sources from next version 0.99.28

Issue #1269, opened 2020-10-19
Wrong bullet characters


Hoffenbar:
**v099.16, Windows**

After switching to other fonts, the bullets in bulleted lists are wrong, e.g. in the toc. The list of bullet characters in the respective configuration dialog looks right instead.

See attached screenshot.
[2020-10-19 23_24_59-Window.zip](https://github.com/giuspen/cherrytree/files/5404847/2020-10-19.23_24_59-Window.zip)

txe:
The reason is that the fonts don't support some unicode symbols. You can change fonts, or change symbols in Preference dialog, or wait for resolving  #1065.

kettenfett:
#1269

txe:
The issue was fixed, the fix will be in 0.99.18

Issue #1245, opened 2020-10-14
[0.99.15] Special Characters for bulleted list and TOC still garbled.


kettenfett:
As of 0.99.15, Special Characters still look garbled. #1133 

I have set **Arial** in Preferences as Rich text font and plain text font.
Also in Preferences under Special Characters I set "•▸•▸•▸" as chars for bulleted list and "▸•◇▪" for TOC.
Then restarted cherrytree  and made a new document.

Result is:
![grafik](https://user-images.githubusercontent.com/23098362/95968295-90b3d380-0e0d-11eb-8a44-f21036e28cf7.png)

Sooo, how is this fallback font thingy coming along?

ciranus:
Which OS ???

ciranus:
Can you confirm with 0.99.**17** ?

> fix visualisation glitches of nodes with anchored widgets (#1205 )

https://www.giuspen.com/cherrytree/#downl

kettenfett:
> 
> 
> Which OS ???
Windows 10 x64




> 
> 
> Can you confirm with 0.99.**17** ?
I tried. It's still the same. Upon extracting the portable archive, rich text font was reset to Cambria Normal, with which chars work. But not any other fonts I tested.

txe:
The issue was fixed, the fix will be in 0.99.18

PR #1278, opened 2020-10-26
adds fallback fonts on Win32

### LINKED ISSUE #1065: C++: Windows Fallback Font / Can't Type Multiple Newlines (interrelated bugs)
In the current C++ master (built from source on Windows), I'm not able to enter more than one blank newline at a time.  To reproduce:

1) Create a new node
2) Type a few characters
3) Press Enter.  At this point, the flashing cursor disappears, but we can assume the cursor is at the start of the 2nd line
4) Press Enter again
4) Type a few more characters

The result is that the two lines of characters are right on top of each other.  The expectation is that there should be a blank line between them.  In other words, it only registered one "enter" press, even if I press "enter" i.e. 2x (or more).

You can also reproduce this by i.e. typing some characters, moving the cursor to the middle of those characters, then pressing Enter a bunch of times.  Only the first press of "enter" registered.

txe:
Resolves #1065 #1269 #1245


txe:
I also checked `cherrytree_0.99.17.0_win64_portable`, it contains the fixed pango, so when I replaced .exe file, fallback fonts worked fine.

PR #1329, opened 2020-11-09
new command to rename emb. files

### LINKED ISSUE #1301: Feature Request: Rename Attachments
It would be great if you could rename file attachments from the context menu (i.e. without needing to save them to the system, delete from the note, rename, then re-attach).

txe:
Resolves #1301
@giuspen , FIY, it adds new translate strings and uses standard 'ct_edit' icon for popup menu item.


giuspen:
Thanks @txe there are already few new strings to translate I'll ask the translator's help

PR #1368, opened 2020-11-24
Instructions to Build/Debug with VS Code container

spasche:
Using a container to build can be useful if you're on Windows and/or want a self-contained build/debug environment.


giuspen:
Thanks!

PR #1371, opened 2020-11-24
New --export_single_file export option

spasche:
--export_single_file can be used to export to a single text or html
file from the command line.
Also prevent "No Node is Selected" popup from appearing during export.


giuspen:
Sorry for delay looking at this, have been busy with a crash, looking at it in the weekend

spasche:
Sure, no worries. It's not urgent for me ;-) Thanks for the feedback.

giuspen:
good stuff, thanks!

PR #1391, opened 2020-12-02
Snap updates to fix build and release 0.99.24

hellsworth:
Since setup.py moved to the pygtk2 dir, the snap build had to be fixed. This build also is based off of the 0.99.24 tag.


giuspen:
Thanks!

PR #1396, opened 2020-12-02
Fix the cherrytree shebang

hellsworth:
It turns out that the snapcraft used on launchpad (which publishes to the snap store) was expanding this variable while my locally built snaps didn't have the problem.

This will fix the 0.99.24 snap runtime.

Once this change is merged, it will build and be available in the store and then to test:
```
$ sudo snap install cherrytree --edge
```

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -107,7 +107,7 @@ parts:
       cd pygtk2
       python2.7 setup.py install --prefix=$SNAPCRAFT_PART_INSTALL/usr -f 
       sed -i 's|^Icon=.*|Icon=${SNAP}/meta/gui/cherrytree.png|' linux/cherrytree.desktop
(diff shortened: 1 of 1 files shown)


giuspen:
Thanks Heather. I see you are still delivering 0.39.x though, do you prefer to wait a couple of more months so that 0.99.x is more stable?

hellsworth:
Well we can at least put 0.99.24 in the edge channel so it could get some testing. I can hold off on releasing to stable though if you think that would be best?

giuspen:
No I think it's safe even though there are various bits that need to be fixed/polished. My observation is because looking at your changes you are still building/delivering the Pygtk2 stuff, not the Gtkmm3 stuff, am I right?

hellsworth:
>No I think it's safe even though there are various bits that need to be fixed/polished.

Ok

>My observation is because looking at your changes you are still building/delivering the Pygtk2 stuff, not the Gtkmm3 stuff, am I right?
Yes that's right.. I'll work on building the gtkmm3 stuff now though :)

PR #1420, opened 2020-12-11
Update README.md

marianopela:
Had to add `cd cherrytree` after cloning in order for `git submodule update --init` to work as expected and be able to build on macOS

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,6 +166,7 @@ brew install python3 cmake pkg-config gtksourceviewmm3 gnome-icon-theme gspell l
 Get cherrytree source, compile and run:
 ```sh
 git clone https://github.com/giuspen/cherrytree.git
+cd cherrytree
 git submodule update --init
 mkdir cherrytree/build
 cd cherrytree/build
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -166,6 +166,7 @@ brew install python3 cmake pkg-config gtksourceviewmm3 gnome-icon-theme gspell l
 Get cherrytree source, compile and run:
 ```sh
 git clone https://github.com/giuspen/cherrytree.git
+cd cherrytree
 git submodule update --init
 mkdir cherrytree/build
```
giuspen, line 171: I caused the issue adding the git command, but with your cd cherrytree you invalidated the following two, has to become mkdir build && cd build, could you fix that too? May I ask you to do the same for the other places in the file where I added git submodule update --init?

marianopela:
I was over my head and made. I'll delete and redo it clean

README.md
marianopela: Sure I'll add the other ones too

PR #1421, opened 2020-12-11
Fix building instructions on macOS

marianopela:


giuspen:
Thanks!

marianopela:
You're welcome, and thanks for your patience. I seem to be missing too much sleep lately...

Issue #1436, opened 2020-12-17
Build failures on Fedora Rawhide


funnelfiasco:
**Version, Operative system**
CherryTree 0.99.25git44ad7a6 on Fedora Rawhide

**Describe the bug**
Build failures in `src/ct/CMakeFiles/cherrytree_shared.dir/ct_actions_edit.cc.o`. Works in Fedora 32 33. I'm guessing it's gcc 11 related, but I haven't had a chance to dig into it yet.

``` 84%] Building CXX object src/ct/CMakeFiles/cherrytree_shared.dir/ct_actions_edit.cc.o
In file included from /builddir/build/BUILD/cherrytree-44ad7a610c5558ec9b7903a89f07ff9929b9513f/src/ct/ct_dialogs.h:26,
                 from /builddir/build/BUILD/cherrytree-44ad7a610c5558ec9b7903a89f07ff9929b9513f/src/ct/ct_actions.h:26,
                 from /builddir/build/BUILD/cherrytree-44ad7a610c5558ec9b7903a89f07ff9929b9513f/src/ct/ct_actions_edit.cc:24:
/builddir/build/BUILD/cherrytree-44ad7a610c5558ec9b7903a89f07ff9929b9513f/src/ct/ct_misc_utils.h:63:10: error: 'optional' in namespace 'std' does not name a template type
   63 |     std::optional<T> peek() {
      |          ^~~~~~~~
/builddir/build/BUILD/cherrytree-44ad7a610c5558ec9b7903a89f07ff9929b9513f/src/ct/ct_misc_utils.h:36:1: note: 'std::optional' is defined in header '<optional>'; did you forget to '#include <optional>'?
   35 | #include <condition_variable>
  +++ |+#include <optional>
   36 | 
In file included from /builddir/build/BUILD/cherrytree-44ad7a610c5558ec9b7903a89f07ff9929b9513f/src/ct/ct_actions.h:29,
                 from /builddir/build/BUILD/cherrytree-44ad7a610c5558ec9b7903a89f07ff9929b9513f/src/ct/ct_actions_edit.cc:24:
/builddir/build/BUILD/cherrytree-44ad7a610c5558ec9b7903a89f07ff9929b9513f/src/ct/ct_table.h:83:34: error: 'optional' in namespace 'std' does not name a template type
   83 |     int get_col_width(const std::optional<size_t> optColIdx = std::nullopt) const {
      |                                  ^~~~~~~~
[...]

PR #1450, opened 2020-12-22
More GCC11 fixes

funnelfiasco:
Add one more missing `#include <optional>` to make GCC 11 happy.

This is an additional (and final, it appears) fix for #1436.


giuspen:
Thanks!

Issue #1260, opened 2020-10-18
Mac HotKey "Command" (⌘) button support


NSTikhomirov:
**Is your feature request related to a problem? Please describe.**
The Command key, ⌘, also `Meta` in gtk, is a modifier key present on Apple keyboards. Mac users are used to the Command modifier instead of Control.

**Describe the solution you'd like**
GTK + has a "<Primary>" modifier in place of "<Control>". About this modifier written at https://unix.stackexchange.com/questions/168505/what-is-primary-for-a-key

txe:
So, from docs, "Primary" means Meta on Apple and Ctrl everywhere else, it's recommended over the use of Ctrl. In our case, it can affect:
1. default shortcuts
2. user shortcuts already saved in the config file
3. shortcut table in Preferences and shortcut edit dialog
4. build-in shortcut e.g. navigating on the node panel

For MacOs users
1. we can use `<Primary>` instead of `<Ctrl>` in the source, not a big deal
2. nothing to do
3. show `<Primary>` in the table. Add additional button `<Primary>` next to `<Ctrl>` so user can switch between them. Also, we can add a flag in the Pref dialog, so `<Primary>` can work as `<Ctrl>` like in other OS
4. somehow fix build-in shortcuts and use `<Primary>` instead of `<Ctrl>`. Additionally,  we can display these shortcuts somewhere, so users can configure them.

For other users.
1. we can use `<Primary>` and it will work as `<Ctrl>`
2. nothing to do
3. If we show `<Primary>, will it be confusing for uses? Also in this case, we should show `<Primary>` next to `<Ctrl>` button (though it will do the same thing, but users can sync between OS)
4. nothing to do

Do I miss something?

NSTikhomirov:
> So, from docs, "Primary" means Meta on Apple and Ctrl everywhere else, it's recommended over the use of Ctrl. In our case, it can affect:
> 
> 1. default shortcuts
> 2. user shortcuts already saved in the config file
> 3. shortcut table in Preferences and shortcut edit dialog
> 4. build-in shortcut e.g. navigating on the node panel
> 
> For MacOs users
> 
> 1. we can use `<Primary>` instead of `<Ctrl>` in the source, not a big deal
> 2. nothing to do
> 3. show `<Primary>` in the table. Add additional button `<Primary>` next to `<Ctrl>` so user can switch between them. Also, we can add a flag in the Pref dialog, so `<Primary>` can work as `<Ctrl>` like in other OS
[...]

PR #1467, opened 2021-01-03
Add meta modifier shortcut key support

cstahly:
Add ability to use meta shortcuts, primarily for the Command key on MacOS (now with stripping the modifier from the keyboard code in the correct place, and not touching the python code)


giuspen:
Thanks!

NSTikhomirov:
#1260

PR #1488, opened 2021-01-14
Update README.md - Added build description

Nlar:
Added build description for Gentoo.  Used repo supplied ./build.sh instead of cmake commands.  ./build.sh handles the need to use mkdir


giuspen:
Thanks!

PR #1491, opened 2021-01-17
Add gnome-icon-theme dependency

mdirik:
Add gnome-icon-theme dependency to Ubuntu build instructions.

`cherrytree` launch aborts with `gtk-icon-theme-error-quark` exception when gnome-icon-theme isn't installed in a chroot environment.


giuspen:
Thanks!

PR #1523, opened 2021-02-17
Update sv.po

eson57:
Update Swedish translation


giuspen:
Thanks I merged manually

PR #1542, opened 2021-03-01
Update ct_const.h

pedroflor:
Added new icons entries into the arrary: freebsd, tux, windows & www
Also into the icons directory.


giuspen:
I prefer to add the icons myself at once after evaluation, please just provide the icons and I'll assign the ids

PR #1543, opened 2021-03-01
Create ct_skull.svg

pedroflor:
I couldn't upload files, but I was able to create a new one SVG from their source.


giuspen:
thanks I downloaded from your zip folder, going to add all together

PR #1566, opened 2021-03-18
Correct the FSF's postal address

funnelfiasco:
This is flagged by package review tools in Fedora and Debian.

See, e.g, https://lintian.debian.org/tags/old-fsf-address-in-copyright-file.html

Signed-off-by: Ben Cotton <bcotton@fedoraproject.org>


giuspen:
thanks!

PR #1599, opened 2021-04-09
Updates files ru.po and kk_KZ.po

Camilot55:
Updates files ru.po and kk_KZ.po


Camilot55:
Fixed the wrong encoding. 
There is also an incorrect encoding in the cherrytree.pot file, please check it.

Camilot55:
Sorry, I checked, everything is fine, you can see that when downloading the repository, something happens and the encoding in the *.po files changes. 
I checked only my translations and found this encoding error. With a more thorough check.

giuspen:
merged manually, thanks Viktor

Issue #1704, opened 2021-05-27
Selected Tree Node Is Not Visible On Start-Up  


gh9000:
**Version, Operative system**
0.99.37, Linux: Arch Linux, Desktop Environment = Cinnamon 4.8.6-1

**Describe the bug**
When the Cherrytree preference, “Miscellaneous -> System Tray -> Start Minimized in the System Tray", is enabled, the selected Tree Node does not appear in the visible area of the TreeView Panel. Scrolling is required to bring the selected Node into view. This only occurs on Cherrytree startup. Also (and obviously), this behavior only occurs if the selected Tree Node is further down in the TreeView Panel than the height of the Panel, i.e., if the panel can show 50 Nodes based on the height of the panel, and any Node between 1 and 50 was selected, no scrolling would be required. However, any selected Node from 51 and down will be invisible until you scroll to find it. 

Here are a few more observations about this issue:
- With the original code, when preference, “Miscellaneous -> System Tray -> Start Minimized in the System Tray", is disabled, it appears that the code (the 'present()' function @ line 150 in src/ct_main_win.cc) positions the selected Node at the top border of the TreeView panel.
- With the original code, when preference, “Miscellaneous -> System Tray -> Start Minimized in the System Tray", is enabled, it appears that the code (the ‘set_visible(false)’ function @ line 147 in src/ct_main_win.cc ) causes Cherrytree to open with the first Node at the top border of the TreeView panel, but selects the correct Node. The problem is that the selected Node is more likely-than-not to be scrolled out-of-view.
- With the modified code, when preference, “Miscellaneous -> System Tray -> Start Minimized in the System Tray", is disabled, behavior is the same as with the original code.
- With the modified code, when preference, “Miscellaneous -> System Tray -> Start Minimized in the System Tray", is enabled, it appears that the code ( ‘present(); set_visible(false)’ ) causes Cherrytree to open with the selected Node approximately in the middle of the TreeView panel, but always visible, nevertheless.

Keep in mind that while I did not notice any unexpected side-effects during limited use of Cherrytree with the patch applied, the developer(s) may recognize potential issues without even applying them.

[...]

Issue #1703, opened 2021-05-27
StatusIcon Does Not Appear In Tray


gh9000:
**Version, Operative system**
0.99.37, Linux: Arch Linux, Desktop Environment = Cinnamon 4.8.6-1

**Describe the bug**
**Issue 1** - Cherrytree's system-tray icon on first start-up does not appear, but leaves a blank space where the icon should have been in the system-tray (clicking the blank area does nothing). Restarting the Desktop Environment (Cinnamon) or logging-out-then-in, will fix the icon. Sometimes, playing with the DE's  panel height will cause the icon to appear, but it may only partially render (e.g. the leaves of the cherry may be cut off even though the DE's panel is tall enough to accommodate the icon). This occurs whether the Cherrytree preference, “Miscellaneous -> System Tray -> Start Minimized in the System Tray", is enabled or disabled.

**Issue 2** - When starting Cherrytree from the command-line, the following error message appears:
**[2021-05-26 12:29:37.039] [gtk] [critical] gdk_window_thaw_toplevel_updates: assertion 'window->update_and_descendants_freeze_count > 0' failed**

Issue 1 and 2 seem to be related. At least the source of the issues seems to be from the same IF statement located @ lines 108-117 in file ‘src/ct_app.cc’. 

I tried multiple variations of commenting lines (110-113 in src/ct_app.cc) to see what worked, and the results were:
- commenting line 110 ('_rStatusIcon->set_visible(false);') fixed the icon issue with no apparent or unexpected side-effects, but the error message persisted.
- commenting any of lines 111-113 WITHOUT commenting line 110 fixed neither the icon nor error message issue.
- commenting line 110 while commenting any of lines 112-113 fixed  the icon but not the error message issue.
- commenting line 110 while commenting line 111 fixed  the icon AND fixed the error message issue. Commenting additional lines (112-113) did not modify this desired behavior.
- the ‘expected’ side-effects of commenting any of lines 111-113 would be to modify the attributes of the status icon, i.e. name (line 111), title (line 112), and tooltip (line 113) will be left unset.

Keep in mind that while I did not notice any unexpected side-effects during limited use of Cherrytree with the patch applied, the developer(s) may recognize potential issues without even applying them.  Finally, the modified code for Issue 1 & 2 affects the status icon's name attribute at the very least, something that the developer(s) set with intent and may not want to leave unset.

[...]

PR #1707, opened 2021-05-29
bugfix - status icon (missing / bad render) #1703 and tree node visibility #1704

gh9000:
Pull request to address Cherrytree Issue #1703 & #1704.

### DIFF
--- a/src/ct/ct_app.cc
+++ b/src/ct/ct_app.cc
@@ -107,8 +107,12 @@ void CtApp::_on_startup()
 
     if (not _no_gui) {
         _rStatusIcon = Gtk::StatusIcon::create(CtConst::APP_NAME);
-        _rStatusIcon->set_visible(false);
-        _rStatusIcon->set_name(CtConst::APP_NAME);
+        /* The following two lines of original code, when active, cause the
+           StatusIcon to fail to properly appear in the system tray. A blank
+           placeholder, or sometimes a poorly rendered icon, appear in its 
+           place. */
+        //_rStatusIcon->set_visible(false);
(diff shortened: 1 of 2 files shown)


giuspen:
Thanks @gh9000

giuspen:
the problem is that now I have the system tray icon out even if I configured it off

giuspen:
I'm going to sort it out anyway no worries

gh9000:
I think I just fixed it, but I want to do a proper build outside VSCodium and install to test.

Based on a quick test, it looks like simply adding code to explicittly set the StatusIcon visibility to false works:

        // show status icon if it's needed and also check if systray exists
        if (_pCtConfig->systrayOn) {
            _pGtkStatusIcon->set_visible(true);
        }
        else {                                                              // New code
            _pGtkStatusIcon->set_visible(false);         // New code
        }                                                                     // New code 
    }

giuspen:
Yes but this way it blinks on the systray at the start which I would like to avoid. I'm thinking instead to slightly complicate the code to do not instantiate at all the status icon unless needed. The pointer is null if not yet instantiated and that will be handled. It will still never hide the systray when instantiating which I understand causes issues

giuspen:
Leave this to me and don't worry about the complication I appreciate you got rid of that critical Gtk error message

gh9000:
My ego is not hurt at all. 

I fully expected you to find better ways to handle the issues. I'm not a Gtk / C++ guy at all. I just thought I'd dig around in the code and give some food for thought to potential fixes rather than just throwing up  a bug report.

I also learned something about interacting with GitHub projects. I've only used  git locally before now. 

Thank you for looking into it so quickly.

Issue #1024, opened 2020-07-19
Feature Request: Make the app scalable for phosh / mobile linux


petrisch:
Its maybe out of the scope, but wouldn't it be nice to have cherrytree on a phone?
Since mobile linux is in its early stages this can be a long run to go.

Here is how fare I came:
[cherrytree on the pinephone ](https://petrisch2.duckdns.org/apps/cms_pico/pico/pinephone/cherrytree)
I guess the main window has its fix sice, that wouldn't scale.

The shortcuts to disable the treeview (F9) doesn't work, but that could be a thing with squeekboard not giving the input.
So whats your thoughts?

txe:
0.37 version uses Gtk2, you can try 0.99 version with Gtk3 (https://www.giuspen.com/cherrytree/#testing) It should work better with wayland.

petrisch:
Thanks for the explanation. I will have to build from source in this case, since the packages are only amd64 and not arm. Will be a Challenge for me. I know there is a flatpak version but not from the testing branch. Or is that to find anywhere?

giuspen:
The flatpak is currently for Pygtk2 but shortly will switch to the Gtkmm3 version

petrisch:
@giuspen I just updated my website. It works on the pinephone, and i can use it. If you want me to test anything on it just let me know. I'm not a dev, but I will help with all i can. Thanks so far!

giuspen:
I'm happy the gtkmm3 version runs on your phone, note that there is also the arm64 version from the ppa. If you find issues report here ;)

DiagonalArg:
I just put up an issue requesting this (closed as identical).  I do take notes on the phone, though.  Having markdown is sufficient for my purposes.  It would be nice not to have to use two separate note-taking appliances.

poleguy:
I now have a pinephone running debian phosh and installed cherrytree. It works, but the menus are cut off. I'd like to work on fixing the menus for small screens... @petrisch any tips on installing the latest development version to test, modify, report bugs, and submit pull requests?

petrisch:
@poleguy not really, grab the source and go for it :-)

PR #1778, opened 2021-08-20
Added FormFactor Property to .desktop

petrisch:
This adds the line explained on [Linmob.net](https://linmob.net/phosh-0-12-app-drawer/)
It could also be the kde variant.
Also I'm hesitant about the value. If its both Workstation and Mobile, there is no differentiation.
However maybe later this can become handy together with #1024.


giuspen:
Thanks!

PR #1792, opened 2021-09-13
Add Flip Image to Image Properties Window

v31l:
Add Flip Image Functionality to the Image Properties Window. Useful feature that was missing.

### DIFF
--- a/pygtk2/modules/support.py
+++ b/pygtk2/modules/support.py
@@ -1103,20 +1103,28 @@ def dialog_image_handle(father_win, title, original_pixbuf):
     hbox_1.pack_start(scrolledwindow)
     hbox_1.pack_start(button_rotate_90_cw, expand=False)
     hbox_1.set_spacing(2)
+    button_flip_horizontal = gtk.Button()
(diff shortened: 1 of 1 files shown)


pygtk2/modules/support.py
```diff
@@ -1152,6 +1160,12 @@ def on_button_rotate_90_ccw_clicked(*args):
         img_parms.height = img_parms.width
         img_parms.width = new_width
         image_load_into_dialog()
+    def on_button_flip_horizontal_clicked(*args):
+        img_parms.original_pixbuf = img_parms.original_pixbuf.flip(true)
+        image_load_into_dialog()
+    def on_button_flip_vertical_clicked(*args):
+        img_parms.original_pixbuf = img_parms.original_pixbuf.false(false)
```
giuspen, line 1167: thanks for your work first of all, this above is flip(false) I believe, not false(false) right?

v31l: Yes. You are correct. It was a mistype.

giuspen:
nice work, thanks

PR #1896, opened 2021-12-16
Updated translations, added new changes

Camilot55:
Updated translations, added new changes

### DIFF
--- a/src/ct/ct_actions_find.cc
+++ b/src/ct/ct_actions_find.cc
@@ -139,7 +139,7 @@ void CtActions::find_in_multiple_nodes()
         if (not entry_predefined_text.empty()) {
             _s_options.str_find = entry_predefined_text;
         }
-        title = _s_state.replace_active ? _("Replace in Multiple Nodes") : _("Find in Multiple Nodes");
+        title = _s_state.replace_active ? _("Replace in All Nodes...") : _("Search in All Nodes...");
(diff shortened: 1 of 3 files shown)


Camilot55:
Updated the name of the dialog box so that it would not work like this (see the screenshot

)

Camilot55:
I suggest using dialog names only for dialog windows.

giuspen:
Hi Viktor, the cherrytree.pot file (parent of all the *.po files) has to be updated first, I will do it freezing the translatable strings soon but before I have to add few more changes, in particular I'm fixing duplicated Alt accelerators for the top menus and that will cause few more updates

src/ct/ct_actions_find.cc
```diff
@@ -139,7 +139,7 @@ void CtActions::find_in_multiple_nodes()
         if (not entry_predefined_text.empty()) {
             _s_options.str_find = entry_predefined_text;
         }
-        title = _s_state.replace_active ? _("Replace in Multiple Nodes") : _("Find in Multiple Nodes");
+        title = _s_state.replace_active ? _("Replace in All Nodes...") : _("Search in All Nodes...");
```
giuspen, line 142: this change is incorrect, because you can select either all the nodes or only the selected node and subnodes in the dialog

Camilot55:
Hi Giuseppe. I have a suggestion for optimizing search and replace. I will try to explain it more clearly in English. It sounds clearer to me in Russian.

PR #1901, opened 2021-12-17
Update ct_menu_actions.cc

Mr-Update:
Shift removed.
The pressed keys Shift+Ctrl+> result in Ctrl+>.
The pressed keys Shift+Ctrl+< also result in Ctrl+>.


giuspen:
Ok, thanks

Issue #2112, opened 2022-09-03
MacPorts install of cherrytree


DrVoIP:
Mac Monterey Ver 12
Macports installed
Installed CherryTree with no errors reported
unable to start the application?

giuspen:
I'm not familiar with Macports but if you run cherrytree from a terminal error messages may be printed to help understand what's wrong

afield1235:
Hello,
A Ticket was opened on this with the MacPorts project: https://trac.macports.org/ticket/65743

They are going to likely create an upstream ticket that this ticket could be linked to once they have all the info.

mascguy:
I'm the MacPorts maintainer for CherryTree, and we're seeing the following crash at startup. This is occurring consistently, across a range of macOS releases:

```
Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   cherrytree                    	       0x10fec4f0b void spdlog::logger::log_<fs::path const&>(spdlog::source_loc, spdlog::level::level_enum, fmt::v9::basic_string_view<char>, fs::path const&) + 43
1   cherrytree                    	       0x10fe0ecf1 CtConfig::_load_from_file() + 327
2   cherrytree                    	       0x10fe0de9f CtConfig::CtConfig() + 3099
3   cherrytree                    	       0x10fde91c5 CtApp::_on_startup() + 235
4   cherrytree                    	       0x10fde9bf0 CtApp::on_activate() + 28
5   libgiomm-2.4.1.dylib          	       0x110e5cc2c Gio::Application_Class::activate_callback(_GApplication*) + 78
6   libgobject-2.0.0.dylib        	       0x111f31074 _g_closure_invoke_va + 201
7   libgobject-2.0.0.dylib        	       0x111f45ff7 g_signal_emit_valist + 988
8   libgobject-2.0.0.dylib        	       0x111f46a70 g_signal_emit + 120
9   libgio-2.0.0.dylib            	       0x1114e759f g_application_real_local_command_line + 1309
10  libgiomm-2.4.1.dylib          	       0x110e5dcaa Gio::Application::local_command_line_vfunc(char**&, int&) + 76
11  libgiomm-2.4.1.dylib          	       0x110e5c84a Gio::Application_Class::local_command_line_vfunc_callback(_GApplication*, char***, int*) + 94
12  libgio-2.0.0.dylib            	       0x1114e5950 g_application_run + 347
13  cherrytree                    	       0x10fd7c1db main + 2103
[...]

PR #2118, opened 2022-09-14
legacy_canonicalize_filename: manage empty filename

kencu:
this function crashes if the filename is an empty string just return an empty string if it comes in as one
see: https://github.com/giuspen/cherrytree/issues/2112


giuspen:
Thanks!

PR #2135, opened 2022-10-15
Fix compilation for MSYS2 Clang64

F3real:
Fixes two issues I've found when trying to compile using MSYS2 clang64
```
C:/msys64/clang64/include/libxml++-2.6/libxml++/validators/xsdvalidator.h:149:8: error: no template named 'auto_ptr' in namespace 'std'
  std::auto_ptr<Impl> pimpl_;
  ~~~~~^
6 errors generated.
```
and

```
ld.lld: error: duplicate symbol: MyFree
>>> defined at C:/msys64/home/user/cherrytree/src/7za/C/Alloc.c:82
>>>            lib7za_static.a(Alloc.c.obj)
>>> defined at SETUPAPI.dll
```


giuspen:
Thanks!

PR #2204, opened 2023-01-25
fixed crash when creating lightweight table

Alex78-cyber:
When I tried create a lightweight table, it crashed. It turned out that the data type of table cell is not right for lightweight table. 
I just make a simple fix. I hope it can help.


giuspen:
Thank you very much I'm glad somebody is using the latest from GitHub especially when close to a release!

PR #2231, opened 2023-02-26
Added missing dependences to Dockerfile for dev container

jvansciver:
Docker build environment provided by VSCode devContainers is currently broken due to missing ubuntu dependencies declared in the Dockerfile.

Also changed to "ms-vscode.cpptools-extension-pack" to be available as an extension in the docker build environment.


giuspen:
Thanks!

PR #2274, opened 2023-05-13
Fix building with fmt 10.

FtZPetruska:
Build is failing with fmt 10 due to changes with how enums are treated.

Specifically, an enum now need to have a `format_as` function, or be casted to its underlying type:
```
Removed deprecated implicit conversions for enums and conversions to primitive types for compatibility with std::format and to prevent potential ODR violations. Use format_as instead.
```


giuspen:
Thanks!

PR #2293, opened 2023-06-14
New langs

Camilot55:
New langs


giuspen:
Good job, thanks Viktor

PR #2312, opened 2023-07-15
100% ru.po and kk_KZ.po and kk_LA.po

Camilot55:
New langs


giuspen:
many thanks Victor  I merged manually

PR #2314, opened 2023-07-17
swap all hotkeys to lowercase in menu

GrylledCheez:
The uppercase letters can be misleading if they have both an upper and lowercase counterpart.


giuspen:
ok for me, thanks

PR #2321, opened 2023-07-24
Snap updates needed for 1.0.0

hellsworth:
- Remove unnecessary libraries
 - Add libxml2-utils dependency
 - Set LD_LIBRARY_PATH without overwriting it


giuspen:
Thanks!

PR #2416, opened 2023-12-23
Update debian control file with new build dependencies 

DennisRasey:
Adding new build dependencies to Debian/Ubuntu control files

> file
> libxml2-utils

Additionally adding following to Ubuntu 18.04 control file.
These were missing in 18.04 but present in 20.04 & 22.10 control files.
Verified both packages are available in 18.04

> libfmt-dev
> libspdlog-dev

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -53,7 +53,7 @@ gdb ./build/cherrytree
 ## Building Cherrytree on Debian (+Ubuntu/Linux Mint)
 Install dependencies:
 ```sh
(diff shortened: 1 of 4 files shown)


giuspen:
Actually 
libfmt-dev
libspdlog-dev
are missing on purpose on 18.04 as the version of the library there is too old so it's building with the bundled spdlog library which has embedded a version of fmt.
Could you also merge from the latest master into your branch as there is a conflict

PR #2457, opened 2024-03-14
100% ru.po and kk_KZ.po and kk_LA.po

Camilot55:
Lang update


giuspen:
Great job, thank you Viktor

Issue #2516, opened 2024-05-30
Duplicate text in File menu (Portuguese translation)


jmg99:
**Version, Operative system**
1.1.2, Windows

**Describe the bug**
Duplicate "Open File", "Open File" menu items, instead of "Open File" and "Open Folder".

**To Reproduce**
1. Run Cherrytree with the Portuguese translation.
2. Click on the "Ficheiro" menu.

**Screenshots**
![image](https://github.com/giuspen/cherrytree/assets/17392738/8dfb921f-011c-4932-abdf-df7493e153c8)

PR #2517, opened 2024-05-30
Fix duplicate "Open File" in PT translation

jmg99:
Now has both "Open File" and "Open Folder"
Fixes giuspen/cherrytree#2516


giuspen:
Thanks!

Issue #1831, opened 2021-10-16
Search Results Ordered By Recently-Accessed


metal450:
One more search request ;)

When doing find-in-all, it would be great if the search results dialog could be ordered by most-recently accessed node. That way, one can avoid scrolling down through very old nodes that potentially haven't been accessed for years, & more commonly-viewed content will naturally find its way to the top. This is how i.e. OneNote works, too :)

Ricent82:
+1 to this. Specifically, ordered by the most recent one that was accessed *from the search dialog*.

I find myself searching for the same notes very often, but they always appear at the bottom of the search results, would be much faster if the "often visited from search" nodes would be remembered & ordered first.

PR #2552, opened 2024-08-10
(feat): Adding new quick-select dialog

ericguin:
This adds a new quick-select dialog much akin to the Command Palette, except instead of selecting commands, it fuzzy finds on node name/path.

This somewhat improves the UX, in my opinion, for quickly switching nodes. There is the `Ctrl+T` shortcut for "Find in Multiple Nodes...", but that search box does not have suggestions or live filtering.

The inspiration for this feature comes from the `Ctrl+P` default binding in VS Code where you can search for a file very quickly and switch to it. I have put a default binding of `Ctrl+Alt+P` in, but that is a very rough "what I could find that wasn't already used" kind of map and I am very open to changing it.

Also note, if there is some reason this feature hasn't been in CherryTree I am happy to abandon the PR. I just think it might make a really cool addition.

### DIFF
--- a/src/ct/ct_actions_file.cc
+++ b/src/ct/ct_actions_file.cc
@@ -387,6 +387,14 @@ void CtActions::preferences_export()
     _pCtConfig->write_to_file(filepath);
 }
 
+void CtActions::command_selnode()
+{
+    auto id = CtDialogs::dialog_selnode(_pCtMainWin);
+    auto node_iter = _pCtMainWin->get_tree_store().get_node_from_node_id(id);
+    _pCtMainWin->get_tree_view().set_cursor_safe(node_iter);
+    _pCtMainWin->get_text_view().grab_focus();
+}
+
 void CtActions::command_palette()
 {
     std::string id = CtDialogs::dialog_palette(_pCtMainWin);
--- a/src/ct/ct_actions_file.cc
+++ b/src/ct/ct_actions_file.cc
@@ -390,9 +390,12 @@ void CtActions::preferences_export()
 void CtActions::command_selnode()
 {
     auto id = CtDialogs::dialog_selnode(_pCtMainWin);
-    auto node_iter = _pCtMainWin->get_tree_store().get_node_from_node_id(id);
-    _pCtMainWin->get_tree_view().set_cursor_safe(node_iter);
-    _pCtMainWin->get_text_view().grab_focus();
+    if (id != -1)
+    {
+        auto node_iter = _pCtMainWin->get_tree_store().get_node_from_node_id(id);
+        _pCtMainWin->get_tree_view().set_cursor_safe(node_iter);
+        _pCtMainWin->get_text_view().grab_focus();
+    }
 }
 
 void CtActions::command_palette()
--- a/src/ct/ct_dialogs_sel_node.cc
+++ b/src/ct/ct_dialogs_sel_node.cc
@@ -64,11 +64,10 @@ gint64 CtDialogs::dialog_selnode(CtMainWin* pCtMainWin)
 
     int order_cnt = 0;
 
-    treeStore.get_store()->foreach([&](const Gtk::TreeModel::Path& path, const Gtk::TreeModel::iterator& iter)
+    treeStore.get_store()->foreach_iter([&](const Gtk::TreeModel::iterator& iter)
     {
         auto ctit = treeStore.to_ct_tree_iter(iter);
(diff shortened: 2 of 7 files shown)


src/ct/ct_dialogs_sel_node.cc
```diff
@@ -0,0 +1,242 @@
...
+    auto list_store = Gtk::ListStore::create(columns);
+    auto& treeStore = pCtMainWin->get_tree_store();
+
+    int order_cnt = 0;
+
+    treeStore.get_store()->foreach([&](const Gtk::TreeModel::Path& path, const Gtk::TreeModel::iterator& iter)
+    {
+        auto ctit = treeStore.to_ct_tree_iter(iter);
+        auto full_path = CtMiscUtil::get_node_hierarchical_name(ctit, " / ", false);
+        spdlog::debug("Node path for {} in ints is: {}", ctit.get_node_name(), path.to_string());
```
ericguin, line 71: Left in my debug on accident

src/ct/ct_actions_file.cc
```diff
@@ -387,6 +387,14 @@ void CtActions::preferences_export()
     _pCtConfig->write_to_file(filepath);
 }
 
+void CtActions::command_selnode()
+{
+    auto id = CtDialogs::dialog_selnode(_pCtMainWin);
+    auto node_iter = _pCtMainWin->get_tree_store().get_node_from_node_id(id);
```
giuspen, line 393: as you can return an invalid id (-1) for example if the user presses cancel on the dialog, you have to check that the returned node_iter is valid `if (node_iter )` or set_cursor_safe() with a bad input will crash

src/ct/ct_dialogs_sel_node.cc
```diff
@@ -0,0 +1,241 @@
+/*
+ * ct_dialogs_cmd_palette.cc
```
giuspen, line 2: file name to fix after copy/paste

src/ct/ct_menu_actions.cc
```diff
@@ -293,6 +293,8 @@ void CtMenu::init_actions(CtActions* pActions)
         _("Replace Tabs with Spaces"), sigc::mem_fun(*pActions, &CtActions::replace_tabs_with_spaces)});
     _actions.push_back(CtMenuAction{tools_cat, "command_palette", "ct_execute", _("_Command Palette..."), KB_CONTROL+KB_SHIFT+"p",
         _("Command Palette"), sigc::mem_fun(*pActions, &CtActions::command_palette)});
+    _actions.push_back(CtMenuAction{tools_cat, "select_node", "ct_execute", _("_Select Node..."), KB_CONTROL+KB_ALT+"p",
+        _("Select Node"), sigc::mem_fun(*pActions, &CtActions::command_selnode)});
```
giuspen, line 297: there is an issue with the windows gtk library so that when the user presses AltGr it detects Ctrl+Alt (causing unwanted clashes on non english keyboard layouts characters input), because of this we cannot use any Ctrl+Alt keyboard shortcuts. You can leave it without a shortcut for now so that the user individually can assign one, I will look for another free one

giuspen (changes requested):
Thanks for sharing your work, please see my review comments

src/ct/ct_actions_file.cc
ericguin: D'oh... Completely slipped my mind on that one. Even added the -1 *for this very purpose*

ericguin:
Thank you for the prompt review! Really loving this software, there's something about it that is finally making it not just easy but fun to finally start organizing my life. Much appreciated.

giuspen:
very nice work @ericguin thanks again for sharing it, I will look for a suitable default keyboard shortcut

metal450:
This is amazing. A few notes/feedback:

* Could we have this respect the search exclusions? Ideally with an "Override exclusions" toggle, similar to the main search dialog. In my case this is a top issue, as i.e. 75%+ of nodes are in an excluded "Done" folder - but from this search, they're all included,
* The dialog pops up off-center on the right side of the screen, and then as it populates, becomes even wider, eventually clipping off the right of the screen. This requires manually moving it to the left each time, if you want to see the full names in the results. It would be good to recenter as it populates & resizes.
* It would be great if the search results could be ordered by most-recently accessed node. That way, one can avoid scrolling down through old nodes that potentially haven't been accessed for awhile, & more commonly-viewed content will naturally find its way to the top. This is how i.e. OneNote works, too. Suggested it awhile ago for the main search, but would be even better here: https://github.com/giuspen/cherrytree/issues/1831

PR #2626, opened 2025-01-04
Add cropping to insert image dialogue

Uks2:
Hi,

This adds the ability to select an area to crop to in the insert image dialogue.

You're obviously welcome to it as is, however:
1. I don't know if dumping a class into a source file like that is the done thing in C++, should it be declared in a header?  I've never worked on a C++ project big enough for that sort of thing to matter!
2. There's currently no indication in the UI that this feature is there or how it relates to the width and height options.  (At the moment, the whole image is scaled according to the width and height values, then cropped according to the selection.)  I'm not sure what the best way to do the UI would be now.
3. I'm still having `load_image_into_dialog` create the preview pixbuf at whatever resolution it sees fit. It then hands that to the `CropImage` widget which gets the crop area relative to the preview resolution.  When you press OK, the `image_handle_dialog` function gets the chosen total image resolution, hands that to the image widget and gets back the final crop area relative to that.  This seems like a rather round about way of doing things.  I'd probably move as much as possibe into the `CropImage` widget class, but how exactly you go about it might depend on the solution to problem 2.

https://github.com/user-attachments/assets/399367cc-0d17-438d-aea4-0c6f8073477d

(I was also having a go a getting cherrytree to automatically insert images generated by run code, but that's a different thing.  The dialogue behaves the same regardless of where the image is coming from.)

### DIFF
--- a/src/ct/ct_dialogs_anch_widg.cc
+++ b/src/ct/ct_dialogs_anch_widg.cc
@@ -97,6 +97,146 @@ Glib::ustring CtDialogs::latex_handle_dialog(CtMainWin* pCtMainWin,
     return Gtk::RESPONSE_ACCEPT == dialog.run() ? rBuffer->get_text() : "";
 }
 
+class CropImage : public Gtk::Image {
+
+public:
+    CropImage(Glib::RefPtr<Gdk::Pixbuf> pixbuf) :
+            Gtk::Image(pixbuf) {
+
+        x = 0;
+        y = 0;
+        w = pixbuf->get_width();
+        h = pixbuf->get_height();
+        moved = false;
+
+        set_has_window(true);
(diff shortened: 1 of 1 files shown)


giuspen:
Many thanks for sharing your work @Uks2 looks great, let me review a bit more in detail and I'll merge in

giuspen:
I find it outstanding, and certainly useful to many, thanks again for sharing your work.
As you also state, I'm thinking about adding some text maybe at the bottom of the image to let the user know that cropping can be performed simply selecting a rectangle on top of the image. Maybe this text can be shown only after a button click - a new button, clearly symbolising cropping, added before or after the width [ ] height [ ] labels/spinbuttons.

PR #2635, opened 2025-01-18
Fix duplication of codebox toolbar callbacks

### LINKED ISSUE #2628: Changing the syntax of a codebox
**Version, Operating system**
CherryTree version: 1.2.0
OS: Ubuntu 20.04.1
Desktop: Gnome

**Describe the bug**
Changing the syntax of a codebox multiple times will increase the amounts of the code box property window exponentially.
As a result, the number of windows you have to close increases each time.

**To Reproduce**
1. Create a codebox. (Shift + Alt + D)
2. Select the Codebox.
3. Right click on the codebox 
4. Click on "Change Codebox Properties". (Shift + Alt + D)
5. Choose any language for syntax highlighting
6. Click "OK" button in the "Automatic Syntax Highlighting" window.
7. Click "OK" button in the "Edit CodeBox" window.
8. Repeat this process several times starting from step 2. 

**Screenshots**
aa-ezgif com-video-to-gif-converter (1)

Uks2:
The `update_toolbar_buttons` function, which is called whenever a codebox is reconfigured or when codebox toolbars are en-/disabled, works by removing all the toolbar buttons and readding them as necessary.  It was using the `_toolbar.append()` function which sets up the buttons' callback at the same time.  Because the ToolbarButton objects are persistent, an extra copy of the callback was being connected each time the button was removed and readded, so Gtk thought we wanted to run it multiple times.

Now, the buttons' callbacks are setup once using
`signal_clicked().connect()` in the codebox constructor and the `update_toolbar_buttons` function uses `_toolbar.insert()` which doesn't affect the callbacks at all.

Fixes #2628


giuspen:
Nice one thanks @Uks2

PR #2908, opened 2026-08-21
Update fr.po

BenJi-TR:
Fixed French message box for tab replacement feature.
We're talking about "tabulations" here and not "onglets".


giuspen:
Merci!

PR #2918, opened 2026-09-04
Port to libxml++5

bgermann:


giuspen:
It's a nice work but libxml++5.0 library is only available in Ubuntu 26.04.
I cannot drop support for older distributions.
I will have a think, maybe adding macros so it is buildable with both is the best solution.

PR #2923, opened 2026-09-09
Reject empty tables during XML parsing

acts-1631:
Table XML is parsed by CtStorageXmlHelper::populate_table_matrix()
before _create_table_from_xml() constructs the table widget. With no
row elements, the parser called tableMatrix.back() on an empty vector.
The clipboard path could also index an unvalidated empty matrix.

Return failure for tables without rows or cells, and handle that result
before table construction or clipboard indexing. This prevents a
crafted document or clipboard payload from crashing CherryTree through
out-of-bounds access while preserving valid tables.


giuspen:
Thanks @acts-1631

